Issue #4, opened 2016-09-25
first time sign-in, did not sign me in


jarradh:
first time signing in, retuned me to commiteth and i had to press sign in button again.

PR #9, opened 2017-01-15
Fix first-time sign in issue

tpatja:
* db/create-user! now returns a map with the created data instead of a
number, as expected by the client side.
* fixed broken test

Fixes: #4


jarradh:
thanks @tpatja !

Issue #40, opened 2017-03-15
Keep github oauth tokens on the client side


tpatja:
### User Story

As a user, I expect commiteth to respect my privacy and not store my github oauth token on the server.

### Description

*Type*: Enhancement

*Summary*: There should be no need to store github oauth tokens on the server.

### Solution
* after successful redirect to commiteth.com, pass token to client-side and put it in localStorage on the client, avoiding the need to store the token on the server
* each commiteth HTTP API call that uses functionality in the github API requiring the user's token should be passed the token (we use ssl, so no security issue here)

Issue #35, opened 2017-03-14
Progressive permissioning


tpatja:
### User Story

As a user, I only want to grant commiteth permissions that are needed.

### Description

*Type*: Feature

*Summary*: Require only basic permissions when signing up and ask for more permissions when needed.

### Solution

* require only `user:email` when signing up 
* require additionally `admin:repo_hook`,  `repo` and `admin:org_hook` in case the user wants to use the Manage repositories feature (to be a repo admin). This can work so that if the user does not already have the full permission set, there is a button on the "Repositories" tab for granting access.

Sija:
I was off put by requirement for an extended attributes (private repos), and especially giving an admin access. Perhaps some information beforehand why it's needed would help to mitigate the experience, still progressive permissioning seems like a better choice.

jarradh:
We removed it, but then none of the organisation repositories were displayed and was forced to add it back in here https://github.com/status-im/commiteth/commit/a9b6a28dec799d7c2b9b3118a2bf0877b0d2288b 

Any ideas?

tpatja:
Without the `repos` oauth scope, it is not possible to access *any* repository data (organization or private). This is a flaw in Github's oauth permissioning scheme (they should have a scope for read-only repository data) and there is an open ticket about it at their end. For now, the best we can do IMO is progressive permissioning, as described in this issue.

Sija:
That's weird, [wafflo.io](https://waffle.io) manages somehow without it.

PR #46, opened 2017-03-22
Progressive Github permissioning & client-side oauth tokens

tpatja:
* require only user:email oauth scope when signing up
* if user wants to set bounties on repos, request additional oauth
  scopes
* do not store github access tokens on server side and use client-side
  localStorage instead

Fixes: #35
Fixes: #40


Sija:
I'm getting Error 500, while trying to sign-in using GitHub :(

tpatja:
Thank you for the bug report @Sija. A fix is now deployed.

PR #53, opened 2017-04-11
feat(eth): Estimate gas instead of hard coding.

### LINKED ISSUE #19: Estimate gas instead of hard-coding
Currently eth.core/send-transaction (used for contract deployments and multisig wallet operations) uses a hard-coded value for gas. We should make the gas value dynamic by using eth_estimateGas.

coopermaruyama:
Closes #19.

### DIFF
--- a/src/clj/commiteth/eth/core.clj
+++ b/src/clj/commiteth/eth/core.clj
@@ -64,8 +64,7 @@
 
 (defn send-transaction
   [from to value & [params]]
-  ;; todo: estimate gas instead of hardcoding
-  (let [gas (format "0x%x" 2600000)
+  (let [gas (estimate-gas from to value params)
         args (merge params {:from  from
                             :value value
                             :gas   gas})]


tpatja:
Thank you for the PR, @coopermaruyama!

IIRC, some contract deployments were running out of gas if we use `eth.core/estimate-gas`. I could be wrong though, will test this soon.

coopermaruyama:
@tpatja Do you remember if you passed the data? I had the same result at first because I was not sending the `data` param to `estimateGas`. But in that case it would likely always fail, not just "sometimes" as you mentioned. Let me know if you find a failing test case.

tpatja:
@coopermaruyama Sorry about my bad wording, deployments were constantly failing before, not sometimes.

I just tested, and your patch does indeed work. You we're also right about the reason why the estimation was not working before (I had omitted the `params` argument to `estimate-gas`).

Thanks again!

PR #71, opened 2017-06-16
Fix Figwheel version mismatch

foopang:


tpatja:
Thanks for the contribution, @foopang!

PR #72, opened 2017-06-23
ERC20 support to multisig wallet

3esmit:
Changed wallet contract. Added a function that send all tokens and ether in contract to an account, this function can only be called by the wallet itself, using execute function. 
Added also watch(address) method that can be used to add a token sent but not notified to contract, so it will be sent together 
The contract support also supports tokenFallback, approveAndCall and manual approve and call.

Still needs some features: 

- [ ] Update balances of tokens; 
- [ ] Schedule balance-of checks into popular tokens (such as SNT and the ones in market) and call watch in case someone did a direct transfer to it using a token that does not call tokenFallback.
- [ ] Instruct users to use approveAndCall instead of direct transfer, especially when using unpopular ERC20 tokens.     
- [ ] Add a option in payment to watch token before send-all.
- [ ] Review contract changes to see if there are any security issues

**IMPORTANT: ONLY WORKS ON RINKEBY** _to deploy in mainnet a library contract needs to be deployed first_
**BREAKS COMPATIBILITY WITH PAST DEPOLYED CONTRACTS**


3esmit:
For now I will only sum all balances in old balance and compare that to see if there is changes and for top hunters. This will work nice for 'updating' comment.

For a next update to fix top hunters there will need a multiplication factor per token to translate value to ETH. We can use ether.delta values or some API. A new field in database would be needed, that would store the amount of value in ETH of sum of tokens at last deposit. This field should be updated for last time when bounty is paid.

For list of tokens, in this first version we will support only a fixed list of tokens, but in future authorized users should be able to watch custom tokens. The list will be like this https://etherscan.io/tokens, covering all popular tokens.

PR #78, opened 2017-07-28
Improved resiliency to partial webhook payload

jeluard:
An exemple of such payload is the [ping event](https://developer.github.com/webhooks/#ping-event).


jeluard:
Actually it was due to incorrectly setting up `commiteth` as a GitHub App and not an OAuth App. 
Not relevant anymore.

PR #79, opened 2017-07-28
Added OAuth App creation steps

jeluard:


3esmit:
This seems right, but we need also webhooks.

jeluard:
Right. Actually the `webhook` registration is handled by `commiteth` process itself.

tpatja:
Outdated PR, closing

PR #83, opened 2017-08-16
wallet factory, and more organizated contracts

3esmit:


contracts/MultiSigStub.sol
```diff
@@ -0,0 +1,67 @@
+pragma solidity ^0.4.15;
+
+/**
+ * @title MultiSigStub  
+ * Contact that delegates calls to a library to build a full MultiSigWallet that is cheap to create.
```
tpatja: 1) Typo (should be `Contact`). 

2) Perhaps should also mention which library the calls are delegated to (looks like it is a library contract expected to be deployed to address `0x370A93cd1DC15875fF02aa0b952D44Bb3dD905E5`, where is the .sol code for that ?)

3esmit: 2. Thats MultiSigTokenWallet.sol
This address might be different for each network, but I can also deploy the contract in all networks at the same address. 
Notice that now Library contract is actually an working multisig that is initialized after the creation of the contract.

contracts/MultiSigStub.sol
```diff
@@ -0,0 +1,67 @@
...
+            m_data := mload(0x40)
+            mstore(0x40, add(m_data, size))
+        }
+    }
+
+    function _delegatecall(bytes32 m_data, uint size) 
+        private 
+        returns(bytes32 m_result) 
+    {
+        address target = 0x370A93cd1DC15875fF02aa0b952D44Bb3dD905E5; //will be replaced by correct value
```
tpatja: How about passing in the library address in the stub constructor? This would avoid hard-coding.

src/clj/commiteth/eth/multisig_wallet.clj
```diff
@@ -31,6 +46,18 @@
...
+        has-factory-event #(some (fn [topic] (= topic
+                                                    factory-topic))
+                                      (:topics %))
+        factory-event     (first (filter has-factory-event logs))
+        factory-data      (:data factory-event)]
+    (when factory-data
+      (subs factory-data 2 66))))
+      
+      
   (defn send-all
```
tpatja: Extra tabs

src/clj/commiteth/eth/multisig_wallet.clj
```diff
@@ -2,8 +2,23 @@
...
+
+(def factory-topic "0x96b5b9b8a7193304150caccf9b80d150675fa3d6af57761d8d8ef1d6f9a1a909")
+
 (def confirmation-topic "0xc0ba8fe4b176c1714197d43b9cc6bcf797a4a7461c5fe8d0ef6e184ae7601e51")
 
+(defn create-new 
+  [owner1 owner2 required]
+  (eth/execute (eth/eth-account)
+               factory 
+               "0xf8f73808"
```
tpatja: Should use `eth/sig->method-id` here

src/clj/commiteth/eth/multisig_wallet.clj
```diff
@@ -2,8 +2,23 @@
   (:require [commiteth.eth.core :as eth]
             [clojure.tools.logging :as log]))
 
+(def factory "0xbcBc5b8cE5c76Ed477433636926f76897401f838")
+
+(def factory-topic "0x96b5b9b8a7193304150caccf9b80d150675fa3d6af57761d8d8ef1d6f9a1a909")
+
```
tpatja: This topic hex-string looks a bit "magic". Perhaps add a comment explaining what it is?

contracts/MultiSigStub.sol
3esmit: Let this for other issue/pr. It envolves creating another contract and maybe even ENS.

PR #101, opened 2017-10-28
Shared static landing page

jeluard:


tpatja:
Can you move the `static_landing_page` directory directly under the root directory? Otherwise all these files ends up in the uberjar.

Issue #110, opened 2017-10-31
Commiteth web: Add Github logo to landing page


asemiankevich:
### Description

*Type*: Bug

*Summary*: On test https://openbounty.status.im:444/ and https://openbounty.status.im/ Github logo is missing on landing page, please add it. Check https://app.zeplin.io/project/59dd4d45719799b4220a2e49/screen/59dd4d9c8f31b2b46b24f233 screen

https://status-im.slack.com/files/U61UDMS5D/F7SF9L878/github_logo.svg

### Expected behavior

![image](https://user-images.githubusercontent.com/28752642/32214684-bf236e90-be30-11e7-9a64-6fee8dc43f7a.png)

### TEST
![image](https://user-images.githubusercontent.com/28752642/32214345-92b29422-be2f-11e7-98f7-f1b6ca7b46ce.png)

### PROD
![image](https://user-images.githubusercontent.com/28752642/32214361-a1130b5a-be2f-11e7-9e1a-cbc0e36eac27.png)

### Additional Information

* Browser: Chrome

tpatja:
Fixed in `develop`

PR #118, opened 2017-10-31
Bug/GitHub logo amend again 110

goranjovic:
addresses #110 (again)

Fixed build-static-page.sh since it was messing up filepaths when target dir already exists.

Re-run it with github logo asset


tpatja:
I believe this is already fixed by commit `cf77f77333`

goranjovic:
You're quick! Ok, I cancel it then

PR #128, opened 2017-11-01
Update services.clj

andytudhope:
Added hutch and myself to the whitelist for testing and so we can make a video for the wiki/update content there generally.


tpatja:
I will merge this, but next time please use more descriptive commit and PR messages. If it's about modifying the whitelist, eg. "Modify whitelist" is much more helpful when viewing commit history that "Update services.clj" (the commit itself indicates what files were changed). Thanks.

PR #134, opened 2017-11-01
Create getting started doc

annadanchenko:
Will update with prod url and same instructions as in the test doc later


doc/getting-started.md
```diff
@@ -0,0 +1,7 @@
+# Getting Started with Open Bounty
+
+Production version is live at
```
tpatja: Seems incomplete

PR #201, opened 2017-12-30
Notification mechanism for bot accounts ETH balance

### LINKED ISSUE #195: Notification mechanism for bot account's ETH balance
*Type*: Feature

Status Open Bounty uses a bot Ethereum account for deploying bounty contacts and signing payouts. This uses ETH for gas. The bot account's ETH has run out a few times, causing things to break.

We need a way of monitoring the bot account's ETH balance and getting notified in case the balance goes below a threshold value.

Requirements:
* configurable threshold value (default: 0.1 ETH)
* post a message to Slack in case balance is below threshold value. This can be done via a HTTP POST (URL will be provided)

pablodip:
Fixes #195 

Could you please tell me the url for the notification?

### DIFF
--- a/src/clj/commiteth/eth/core.clj
+++ b/src/clj/commiteth/eth/core.clj
@@ -103,7 +103,7 @@
     adjusted-gas))
 
 
-(defn from-wei
+(defn wei->eth
   [wei]
   (/ wei 1000000000000000000))
 
@@ -111,10 +111,9 @@
   [eth]
   (biginteger (* eth 1000000000000000000)))
 
-
 (defn hex->eth
   [hex digits]
-  (->> hex hex->big-integer from-wei double (format (str "%." digits "f"))))
+  (->> hex hex->big-integer wei->eth double (format (str "%." digits "f"))))
 
 (defn get-balance-hex
   [account]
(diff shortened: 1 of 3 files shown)


src/clj/commiteth/scheduler.clj
```diff
@@ -342,7 +342,8 @@
   (do
     (log/debug "run-1-min-interval-tasks" time)
     (run-tasks
-     [update-open-issue-usd-values])
+     [update-open-issue-usd-values
+      eth/check-bot-account-eth-threshold])
```
tpatja: This should run less frequently as the scheduler thread very busy already. Hourly is enough.

src/clj/commiteth/eth/core.clj
```diff
@@ -271,3 +270,11 @@
...
             (empty?)))
           true))))
+
+(defn bot-account-eth-threshold-overtaken? []
+  (let [eth (-> (get-balance-wei (eth-account)) wei->eth)]
+    (< eth (env :eth-account-eth-threshold-warning))))
+
+(defn check-bot-account-eth-threshold []
+  (when (bot-account-eth-threshold-overtaken?)
+    (post "NOTIFICATION_URL")))
```
tpatja: * Notification URL can go to config variable `:slack-webhook-url`. If empty, log/error something and abort startup
* No need to make the threshold configurable, just use 0.2 ETH.
* POST contents can be "URGENT: SOB bot account ETH will run out soon, current balance XXX"

tpatja: I will give you the webhook URL privately.

Issue #204, opened 2018-01-03
Use timbre for logging


dmitryn:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story

As a developer, I want to configure app loggers so that i can debug/monitor app easily.

### Description

I suggest to switch from `clojure.tools.logging` to https://github.com/ptaoussanis/timbre because timbre is much easier to configure.

PR #205, opened 2018-01-03
Suppress stdout logging in dev

dmitryn:
Log to file only (info and debug levels). Allows working in REPL without getting log spam.


dmitryn:
Also, i'd like to switch to timbre in future https://github.com/status-im/open-bounty/issues/204

PR #216, opened 2018-01-15
[FIX #156] Add pagination to Bounties and Activities views

### LINKED ISSUE #156: Paging for "open bounties" and activity views
Currently, if there is a lot of data in the system, the views get very long vertically.

Implement client-side paging to UI on "open bounties" and activity views.

siphiuel:
This adds client-side pagination to Bounties and Activities views.


tpatja:
Looks really good @siphiuel! I will take a closer look and test this tomorrow.

tpatja:
Since we do not currently have an easy way to test PRs and the changes look good, I will merge this to `develop` and we can test together in the test env. Possible changes can be submitted via a new PR.

siphiuel:
Thanks! I think it would be good to do server-side paging as well, what do you think?

tpatja:
IMO server-side paging is something we should do, but is not that urgent. The data is gzipped json and transferred data size is not an issue for now. We can create another issue for that later.

pablodip:
I think there are a couple UX things to improve:

* The cursor in the buttons should be a pointer.
* When moving to a different page the screen should probably be moved to the top of the list.

cc @denis-sharypin

siphiuel:
I will implement these if no objections from @denis-sharypin .

siphiuel:
https://github.com/status-im/open-bounty/pull/224

pablodip:
A few more of UX details:

* The button of next and previous page should probably look disabled when the current page is the first and the last one respectively.
* When changing the tab from bounties to activity and viceversa the current page should probably be restarted, that is, always be 1, and right now it is maintained from previous state.
* The activity list should also probably show at the top the message "Showing X of X".
* The next and previous buttons in active state have the same style as the current page button, and it is a bit confusing, since the next and previous are active, and the current page isn't an active thing thing.
* It would probably help to have some style changes when hovering paging buttons.

cc @denis-sharypin

pablodip:
Just created a new issue for those UX details so that we keep better track of them in an open issue instead of a closed pull request.

https://github.com/status-im/open-bounty/issues/226

PR #224, opened 2018-01-19
[FIX #226] Pagination: add cursor pointer for buttons, scroll to top

### LINKED ISSUE #226: UX details in pagination
- [ ] The cursor in the buttons should be a pointer.
- [ ] When moving to a different page the screen should probably be moved to the top of the list.
- [ ] The button of next and previous page should probably look disabled when the current page is the first and the last one respectively.
- [ ] When changing the tab from bounties to activity and viceversa the current page should probably be restarted, that is, always be 1, and right now it is maintained from previous state.
- [ ] The activity list should also probably show at the top the message "Showing X of X".
- [ ] It would probably help to have some style changes when hovering paging buttons.
- [ ] On Safari and FF clicking on page number redirects to random place on page.
Video: https://drive.google.com/open?id=1ZAHPgvmb1C7tuQmYR9-kx_hPVyHRjDc4
- [ ] Pagination should work on mobile, since right now the design is broken.

cc @denis-sharypin @siphiuel

siphiuel:
This will add cursor pointer when hovering over page navigation buttons. Also, now view will scroll upward when navigating to a different page.

Status: Finished.

### DIFF
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -123,8 +123,9 @@
           [:div
            [draw-items] 
            [:div.page-nav-container
-            [draw-rect :backward]
-            [draw-rect :forward]
+            [:div.page-direction-container
+             [draw-rect :backward]
+             [draw-rect :forward]]
             [:div.page-nav-text [:span (str "Page " page-number " of " page-count)]]
             [draw-page-numbers page-number page-count]]])))
 
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -31,22 +31,9 @@
 (defn issue-url [owner repo number]
   (str "https://github.com/" owner "/" repo "/issues/" number))
 
-(defn scroll-div [container-element]
-  "This is an invisible div that exists
-   for the sole purpose of scrolling the container-element into view
-   when page-number is updated"
-  (let [page-number (rf/subscribe [:page-number])]
-    (r/create-class
-      {:component-did-update (fn [] 
-                               (when @container-element
-                                 (.scrollIntoView @container-element)))
-       :reagent-render (fn []
-                         [:div {:style {:display "none"}}
-                          @page-number])})))
-
 (def items-per-page 15)
 
-(defn draw-page-numbers [page-number page-count]
+(defn draw-page-numbers [page-number page-count container-element]
   "Draw page numbers for the pagination component.
   Inserts ellipsis when list is too long, by default
   max 6 items are allowed"
@@ -56,7 +43,9 @@
                             (if current?
                               {:class "page-num-active"}
                               {:class "grayed-out-page-num"
-                               :on-click #(rf/dispatch [:set-page-number i])})
+                               :on-click #(do 
+                                            (rf/dispatch [:set-page-number i])
+                                            (.scrollIntoView @container-element))})
                             i])
         max-page-nums 6]
     [:div.page-nums-container 
@@ -92,7 +81,8 @@
                                  total-count
                                  page-number 
                                  page-count]} 
-                         draw-item-fn]
+                         draw-item-fn
+                         container-element]
   "Draw data items along with pagination controls"
   (let [draw-items (fn []
                      (into [:div.ui.items]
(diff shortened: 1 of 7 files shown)


pablodip:
Is this finished or a WIP? Make sure to indicate it in the description, please. :)

Regarding the scrolling approach, I think we should do it through React, and it could be more advanced with a library like [react-scroll](https://github.com/fisshy/react-scroll) (which supports animations for example) or more simple by using just [refs](https://reactjs.org/docs/refs-and-the-dom.html). Just not sure how difficult would be to use react-scroll in Cljs.

siphiuel:
@pablodip updated the description.

siphiuel:
As to scrolling approach - not sure really which one to pick. `Element.scrollIntoView` does support animations as well, albeit not on all browsers.

pablodip:
Agree with using `Element.scrollIntoView` for now. The main thing in the approach I think is to talk to the DOM through React: https://reactjs.org/docs/refs-and-the-dom.html

siphiuel:
@pablodip thanks for suggestions! Made some changes, refs provide better locality, so no extra effect handlers needed.

siphiuel:
Refactored refs-related code. There is now an additional component that subscribes to page number changes only.

src/cljs/commiteth/handlers.cljs
```diff
@@ -64,19 +63,15 @@
...
- :set-active-page
- (fn [db [_ page]]
-   (assoc db :page page)))
-
-(reg-event-db
-  :set-bounty-page-number
+  :set-active-page
   (fn [db [_ page]]
-    (assoc db :bounty-page-number page)))
+    (assoc db :page page
```
pablodip: What is `:page` used for?

src/cljs/commiteth/common.cljs
```diff
@@ -30,9 +31,22 @@
 (defn issue-url [owner repo number]
   (str "https://github.com/" owner "/" repo "/issues/" number))
 
+(defn scroll-div [container-element]
```
pablodip: Why exactly do we need this no-op component if we can just perform this behaviour in the list components?

src/cljs/commiteth/handlers.cljs
siphiuel: It's an active page keyword, e.g. `:bounties` or `:activity`.

src/cljs/commiteth/common.cljs
siphiuel: The idea was that list components might get updated during any page content changes, not just page number changes.

siphiuel:
Disabled pointer cursor over active page buttons.

src/cljs/commiteth/handlers.cljs
pablodip: Yeah, sorry, I meant what is the difference between `:page` and `:page-number`, that both are being used here.

src/cljs/commiteth/common.cljs
pablodip: I've been thinking about when the scrolling should happen, and it is a little tricky.

The behaviour we are looking for is for the scrolling to trigger when the user navigates to a different page.

And it should not trigger in some other cases:

* As you said, when the list component updates for other reasons.
* And also when the page number changes for other reasons that are not the user navigating. For example, when changing the page, the page number is restarted, but the scrolling should not happen there. Or when the filters change, the page number is restarted as well, and the scrolling should not happen there either.

In order to get this I think the best thing is to pass the ref of the list to the children components so that the buttons where the clicks occur in the navigator have it, and trigger there the scrolling.

I think that is better because we trigger the scrolling exactly where we want it to happen, that is, when the user interacts with the navigator, and don't try to guess that behaviour from state changes in the page number.

src/cljs/commiteth/handlers.cljs
siphiuel: You're right - naming is slightly bad, `:page` should be `:tab` I guess. So it's confusing.

src/cljs/commiteth/common.cljs
siphiuel: Thanks for explaining! Just pushed an update.

PR #230, opened 2018-01-23
Config support and other little improvements

churik:
In this PR:
1) add config support and config_example.ini
2) added additional check to bounties_page
3) make issue title reusable 
4) cleaned up code a bit (added comments, changed some absolute values to values from config)
5) make local as default environment
6) add cleanup option as variable (not brilliant decision) 
7) exclude config.ini from .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,17 +23,3 @@ profiles.clj
 .idea
 resources/contracts
 node_modules
-
-# OS generated files #
-######################
-.DS_Store
-.DS_Store?
-._*
-.Spotlight-V100
-.Trashes
-ehthumbs.db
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -23,4 +23,17 @@ profiles.clj
 .idea
 resources/contracts
 node_modules
+
+# OS generated files #
+######################
```
pablodip: How about managing this with a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)? :)

churik:
@pablodip done, thanks for review

PR #235, opened 2018-01-25
[FIX #185] Readd bounty label

### LINKED ISSUE #185: Ability to readd bounty (delete and add again)
### User Story

As a user, I want to delete label "bounty" when I have problem with deploying contract, ethereum network congestion or if I changed my mind or added bounty by mistake.

### Description
*Type*: Feature
*Summary*: 
when bounty has 0.00 value, or contract address is freezing in deployment stage, I want to remove bounty label and expect that bounty will disappear from SOB. If I add bounty label again, new bounty should again appear on SOB. I suggest to add this feature for contract which don't have money yet, in order to avoid such situations like in issue #184 

#### Expected behavior
I delete bounty label > bounty disappears from SOB site
I add bounty label again > bounty appears on SOB site and I can send money and use it like form scratch

#### Actual behavior
I delete bounty label > bounty does not disappear from SOB site

It can be one of the stage #174

siphiuel:
# Description
This fix does the following:

1. Listen to `unlabeled` events from GitHub API.
2. If such event is received, find the associated issue in the DB and remove it only if funds are zero (currently it checks the field `value_usd` field). This removes the bounty from SOB.
3. If it is successfully deleted from the DB, post to GitHub API  a request to delete a comment. This will remove a bounty comment on GitHub.
4. No actions on part of the contract itself are done.

# Questions
1. Is `value_usd` the correct field to check against when removing a bounty?
2. Should anything be done with regards to the bounty contract when the parent comment is deleted? If so, what messages should be sent?

**Status**: WIP

### DIFF
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -125,6 +125,15 @@ INSERT INTO issues (repo_id, issue_id, issue_number, title)
                    FROM issues
                    WHERE repo_id = :repo_id AND issue_id = :issue_id);
 
+-- :name remove-issue! :<! :1
+-- :doc removes issue
+DELETE FROM issues
+  WHERE repo_id = :repo_id 
+  AND issue_id = :issue_id
+  AND value_usd = 0.0
+  RETURNING comment_id;
+
+
 -- :name update-commit-sha :<! :1
 -- :doc updates issue with commit_sha
 UPDATE issues
--- a/src/clj/commiteth/bounties.clj
+++ b/src/clj/commiteth/bounties.clj
@@ -58,6 +58,16 @@
       (log/debug "Total issues for repo limit reached " repo " " count)
       (add-bounty-for-issue repo repo-id issue))))
 
+(defn remove-bounty-for-issue [repo repo-id issue]
+  (let [{issue-id     :id
+         issue-number :number} issue
+        removed-issue (issues/remove repo-id issue-id)
+        {owner-address :address
+         owner :owner} (users/get-repo-owner repo-id) ]
+    (log/debug "Removing bounty for issue " repo issue-number "owner address: " owner-address)
+    (if-let [comment-id (:comment_id removed-issue)]
+      (github/remove-deploying-comment owner repo comment-id)
+      (log/debug "Cannot remove Github bounty comment as it has non-zero value"))))
 
 ;; We have a max-limit to ensure people can't add more issues and
 ;; drain bot account until we have economic design in place
--- a/src/clj/commiteth/db/issues.clj
+++ b/src/clj/commiteth/db/issues.clj
@@ -12,6 +12,14 @@
                               :issue_number issue-number
                               :title        issue-title})))
 
+(defn remove
+  "Removes issue"
+  [repo-id issue-id]
+  (jdbc/with-db-connection [con-db *db*]
+    (db/remove-issue! con-db {:repo_id      repo-id
+                              :issue_id     issue-id})))
+
+
 (defn update-commit-sha
   "Updates issue with commit-sha"
   [issue-id commit-sha]
--- a/src/clj/commiteth/github/core.clj
+++ b/src/clj/commiteth/github/core.clj
@@ -275,6 +275,10 @@
     (log/debug "Posting comment to" (str owner "/" repo "/" issue-number) ":" comment)
     (issues/create-comment owner repo issue-number comment (self-auth-params))))
 
+(defn remove-deploying-comment
+  [owner repo comment-id]
+  (issues/delete-comment owner repo comment-id (self-auth-params)))
+
 (defn make-patch-request [end-point positional query]
   (let [{:keys [auth oauth-token]
          :as   query} query
--- a/src/clj/commiteth/routes/webhooks.clj
+++ b/src/clj/commiteth/routes/webhooks.clj
@@ -30,13 +30,6 @@
                    (= type (:event %)))
                  events)))
 
-
-(defn labeled-as-bounty?
-  [action issue]
-  (and
-   (= "labeled" action)
-   (= bounties/label-name (get-in issue [:label :name]))))
-
 (defn find-commit-sha
   [user repo issue-number event-types]
   (log/debug "find-commit-sha" user repo issue-number event-types)
@@ -48,12 +41,14 @@
 
 
 (defn handle-issue-labeled
-  [webhook-payload]
+  [webhook-payload label-added?]
   (log/debug "handle-issue-labeled")
   (let [{issue :issue} webhook-payload
         {repo-id :id
          repo-name :name} (:repository webhook-payload)]
-    (bounties/maybe-add-bounty-for-issue repo-name repo-id issue)))
+    (if label-added?
+      (bounties/maybe-add-bounty-for-issue repo-name repo-id issue)
+      (bounties/remove-bounty-for-issue repo-name repo-id issue))))
 
 (defn handle-issue-closed
   [{{{owner :login} :owner repo :name}   :repository
@@ -212,17 +207,18 @@
   [webhook-payload]
   (when-let [action (:action webhook-payload)]
     (log/debug "handle-issue" action)
-    (when (labeled-as-bounty? action webhook-payload)
-      (handle-issue-labeled webhook-payload))
-    (when (and
-           (= "closed" action)
-           (bounties/has-bounty-label? (:issue webhook-payload)))
-      (handle-issue-closed webhook-payload))
-    (when (= "edited" action)
-      (handle-issue-edited webhook-payload))
-    (when (= "reopened" action)
-      (handle-issue-reopened webhook-payload)))
-  (ok))
+    (case action
+      ("labeled" "unlabeled")
+      (when (= bounties/label-name (get-in webhook-payload [:label :name]))
+        (handle-issue-labeled webhook-payload (= action "labeled")))
+      "closed"
+      (when (bounties/has-bounty-label? (:issue webhook-payload))
+        (handle-issue-closed webhook-payload))
+      "edited"
+      (handle-issue-edited webhook-payload)
+      "reopened"
+      (handle-issue-reopened webhook-payload))
+  (ok)))
 
 (defn enable-repo-2 [repo-id full-repo]
   (log/debug "enable-repo-2" repo-id full-repo)
--- a/src/cljs/commiteth/activity.cljs
+++ b/src/cljs/commiteth/activity.cljs
@@ -58,19 +58,25 @@
 
 
 (defn activity-list [activity-page-data]
-  [:div.ui.container.activity-container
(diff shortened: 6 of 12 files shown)


siphiuel:
Changed the comment deletion logic in the DB so that it moves a deleted record to `issues_archived` table instead of dropping it altogether.

pablodip:
Not sure about the current strategy, since having a similar table creates dependency on the original one, so when we want to modify/split the original table we would also have to do it with the archived table, and in the archived data.

The current goal is to just do not lose information, so I agree with the archive concept, just how about implementing it with a more general approach, for example an `archive` table with four columns, `id`, `type`, `archived_at` and `data`, and data could just be the edn of the data that we want to archive? That way it would be more open to archive anything, and we would not have the dependency with the info that we want to archive, so for example the archive would work with any version of the original data. The archived data would not be queriable from SQL, but we would need to take it from there and query it in Clojure, but it is an archive in the end.

Anyway if we seriously aim to not lose information and query better all data over time, that is, to have a time model for the data, we should probably consider using Datomic in the future.

resources/migrations/20180130164001-issues-archived.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE issues_archived;
```
oskarth: Seems safer to rename it and leave it be for a while, then do cleanup task once it is 100% it works well in both test and production.

src/clj/commiteth/routes/webhooks.clj
```diff
@@ -212,17 +207,18 @@
...
-    (when (and
-           (= "closed" action)
-           (bounties/has-bounty-label? (:issue webhook-payload)))
-      (handle-issue-closed webhook-payload))
-    (when (= "edited" action)
-      (handle-issue-edited webhook-payload))
-    (when (= "reopened" action)
-      (handle-issue-reopened webhook-payload)))
-  (ok))
+    (case action
```
oskarth: default case?

siphiuel: Added.

siphiuel:
Changed the code so that deleted issues go to `archive` table. When deploying new contracts, this table is checked to fetch free contracts that haven't been assigned to any issue. So it fulfills a double purpose: that of an archive and a contract pool.

resources/migrations/20180130164001-issues-archived.down.sql
pablodip: This is just a migration down, so why is the alter table really needed? We are not migrating down in production.

siphiuel: I'm not decided on this. Suggest to keep it as is, might help during testing on develop.

pablodip: It's just because this breaks migration semantics I believe. To me migrations should be able to run up and down several times, and this would break that.

resources/sql/queries.sql
```diff
@@ -125,6 +126,23 @@ INSERT INTO issues (repo_id, issue_id, issue_number, title)
                    FROM issues
                    WHERE repo_id = :repo_id AND issue_id = :issue_id);
 
+-- :name remove-issue! :<! :1
+-- :doc removes issue
+WITH archived AS (
+  INSERT INTO archive(type,data)
```
pablodip: How about describing this in clojure? I think that it is better to keep persistence as simple as possible and to have all of the logic in the the application code.

resources/migrations/20180130164001-issues-archived.down.sql
siphiuel: Got you. Sounds reasonable. Changed to DROP.

resources/sql/queries.sql
```diff
@@ -620,3 +638,15 @@ SELECT * FROM (
   ORDER BY 1 desc
   LIMIT :limit_days) AS a
 ORDER BY a.day ASC;
+
+-- :name contract-from-pool :? :1
+-- :doc return first contract from archive for given owner
+SELECT a.data->>'transaction_hash' as transaction_hash
```
pablodip: Not totally convinced about treating an archive to use application logic there apart from archiving. How about modelling the Ethereum contract in a table, even if only with the params that we need right now (I think that `transaction_hash` and `owner_address`), and we don't need to archive it, but just need an attribute for status?

src/clj/commiteth/bounties.clj
```diff
@@ -58,6 +65,16 @@
       (log/debug "Total issues for repo limit reached " repo " " count)
       (add-bounty-for-issue repo repo-id issue))))
 
+(defn remove-bounty-for-issue [repo repo-id issue]
```
pablodip: Probably better ending with `!`

src/clj/commiteth/db/issues.clj
```diff
@@ -12,6 +12,14 @@
                               :issue_number issue-number
                               :title        issue-title})))
 
+(defn remove
```
pablodip: Probably better ending with `!`

src/clj/commiteth/github/core.clj
```diff
@@ -275,6 +275,10 @@
     (log/debug "Posting comment to" (str owner "/" repo "/" issue-number) ":" comment)
     (issues/create-comment owner repo issue-number comment (self-auth-params))))
 
+(defn remove-deploying-comment
```
pablodip: Probably better ending with `!`

resources/migrations/20180130164001-archive.up.sql
```diff
@@ -0,0 +1,17 @@
...
+
+INSERT INTO data_types(name)
+  VALUES('issue'), 
+        ('issue_comment'), 
+        ('repository'), 
+        ('pull_request'), 
+        ('user');
+
+CREATE TABLE archive (
+  type TEXT REFERENCES data_types(name),
```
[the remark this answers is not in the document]
siphiuel: Great idea, changed to enums.

resources/sql/queries.sql
siphiuel: I think it's simple enough to stay in a query, plus one roundtrip less to DB is a good thing.

src/clj/commiteth/bounties.clj
siphiuel: Yes.

src/clj/commiteth/db/issues.clj
siphiuel: +

src/clj/commiteth/github/core.clj
siphiuel: +

resources/sql/queries.sql
pablodip: @siphiuel Let me understand well the query before elaborating the proposal. Why are we selecting `a.data->>'transaction_hash'` and then filtering for it to be null with `a.data->>'transaction_hash' = i.transaction_hash` and `i.transaction_hash is null`? I might be understanding badly the logic. :S

siphiuel: We use left join condition `issues.transaction_hash=null` so that to pick transaction hashes that are not in use already. It might happen that a record has been removed from `issues` table to the `archive`, and then restored. However, the `archive` table is append-only, so this will result in the same transaction hash being present both in `issues` and in the `archive`, and we want to filter that away.

pablodip: Okay, got it, thanks!

So my proposal was about starting modelling the Ethereum contract, since it is a model with some data in the end. And doing so only with the attributes and behaviours of the contract that we need for this.

For example, we could have a table `ethereum_contract`, with attrs `id`, `owner_address`, `transaction_hash`, `status`. When an issue is archived and has a contract that we can reuse, we can create a contract model there with the status `not-used` (or maybe a better name). Then when looking for available contracts, we search there, and when taking one, we change the status of it to `used`. That way we would avoid querying the archive table for application logic, and would express more directly the status of a contract and its filtering.

siphiuel: Let's handle contract modeling in a separate issue: https://github.com/status-im/open-bounty/issues/267.

siphiuel:
Agree.

martinklepsch:
This seems good but I'm also a little bit worried about introducing a separate table. Wouldn't something like `revoked_at` as suggested by @pablodip be easier to work with?

Thinking about https://github.com/status-im/open-bounty/issues/284 we'd need to 

- read the JSON from the archive and parse that 
- handle old versions as the table schema might change over time but the JSON blob stays the same 

Are there any specific benefits with the separate table approach?

siphiuel:
Good point! Initially I've opted for `archive` simply because it was a less intrusive code change and did not imply changing multiple queries and associated middleware, hence smaller probability of regressions. Smart vocabulary aside, I was plain lazy:)

Still, can think of 2 things in favor of archive approach:
1. We'd like to have a DB model that works in an append-only fashion. No data can be deleted without a trace, so that we can recover what happened to each issue and when (audit trail capability). Currently the DB is append-only with regards to issues and pull requests - data is always appended. The archive can record multiple deletions of a single issue's bounty, while `revoked_at` will record the most recent one.
2. Related to a concern that `archive` serves a double purpose: @pablodip suggested modeling contracts in a `ethereum_contract` table, and changing `issues` to include a relation to `ethereum_contract` (https://github.com/status-im/open-bounty/issues/267). This would relieve `archive` of non-archival related duties.

Actually, the whole audit-trail related thing is important and requires more thought.

pablodip:
To me the reason to not face re-modelling here isn't just about laziness, but about strategy. Introducing new tables or fields requires extra work in some existing things, and so might produce bugs as well. And we already intend to re-model with a different structure. So facing re-modelling here and in the new structure isn't probably needed. Besides, the feature that depend on the archive is an optional feature (re-use a contract) and easily refactored if needed.

martinklepsch:
> Actually, the whole audit-trail related thing is important and requires more thought.

Agree. [Event Sourcing](https://martinfowler.com/eaaDev/EventSourcing.html) may be an interesting approach for this as well and is fundamentally similar to how Datomic works.
Given the complexities around this issue my feeling is that we should probably "ignore it" for now and later on allocate some time to think about how to implement it properly.

pablodip:
> Agree. Event Sourcing may be an interesting approach for this as well and is fundamentally similar to how Datomic works.

I personally don't think they're similar, since Event Sourcing is just about modelling and persisting round domain events, while Datomic is a general-purpose database that adds the time dimension.

churik:
## 1. GH comment is not posted after deploying contract

Steps:
- navigate to repo
- create an issue 
- add `bounty` label
- open https://testing.openbounty.status.im/app
- wait until new bounty will be displayed in 'Open bounties' list
- click on bounty

### Actual result:
GH comment with contract details

### Expected result: 
no GH comment

**Video:** http://take.ms/SMVwj
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

PR #236, opened 2018-01-25
Sorting and filtering for open bounties

### LINKED ISSUE #179: SOB bounty sort/filter option
@arash009 commented on [Tue Nov 28 2017](https://github.com/status-im/ideas/issues/48)

## Preamble

    Idea: #48
    Title: SOB Bounty sort/filter option
    Status: Draft
    Created: 2017-11-28

## Summary
There is no way currently to sort or filter the list of bounties on the page. A sort and filter option would be very useful as our list of active bounties start to add up.

## Vision
As a developer looking for bounties to work on, I would like to be able to search, filter, and sort the bounties available, in order that I can find suitable bounties that I'm interested in without having to scroll through a large list.

## Swarm Participants

- Lead Contributor: 
- Testing & Evaluation: 
- Contributor: 
- Contributor: 
- UX (if relevant): @denis-sharypin

## Requirements
Sort options should include:
- Date
- Bounty value
- Repo/Org

Filter options should include:
- Repo/Org
- Min/Max bounty value
- Bounty type? (bug/feature/security/non-development)
- Programming language
- Date

### Minimum Viable Product

Goal Date:  
Description: 

### Iteration 1..N
Goal Date:  
Description: 

## Copyright
Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

pablodip:
Fixes #179 

Status: Finished

#### External libraries

I was gonna use two external libraries, but they didn't work by just importing them with cljsjs, probably because they are not up to date in cljsjs. So I decided to create a minimum version of the features they provide so that we can have a working version of this feature as soon as possible, and add a bounty to externalize their integration once this pull request is merged. The libraries are these:

* https://github.com/Pomax/react-onclickoutside
* https://github.com/react-component/slider

Due to not being able to use react-slider yet and that the input range in html only supports one input value, I created a minimum version of the filter range that we have (the Value one) that uses two ranges instead of one. It looks like this:

screen shot 2018-01-25 at 10 26 36

Once we integrate react-slider it will look like in the design.

Please, let me know if that is okay.

#### Conventions

I've started using namespace keywords so that we can benefit from them. I've also introduced two naming conventions that help a lot with naming things in keywords:

* Using the `.` to namespace in a keyword name.
* Using the `|` to define types.

I also added the suffix `-view` to functions that are actually views, so that they can be distinguised better.

Let me know what you think about them, and if you don't think they are useful I will remove them.


arash009:
Great suggestion re MVP on slider and bounty the integration. Lets do it!

src/cljs/commiteth/subscriptions.cljs
```diff
@@ -110,16 +112,56 @@
...
+      (into #{"ETH"} token-ids))))
+
+(reg-sub
+  ::filtered-and-sorted-open-bounties
+  :<- [:open-bounties]
+  :<- [::open-bounties-filters]
+  :<- [::open-bounties-sorting-type]
+  (fn [[open-bounties filters sorting-type] _]
+    (cond->> open-bounties
+             filters (ui-model/filter-bounties filters)
```
siphiuel: Maybe use `(some second filters)` instead of `filters`?

src/cljs/commiteth/subscriptions.cljs
```diff
@@ -110,16 +112,56 @@
...
+             sorting-type (ui-model/sort-bounties-by-sorting-type sorting-type)
```
siphiuel: And also check if `sorting-type` is not the default value: `(not= sorting-type ::ui-model/bounty-sorting-type|most-recent)`.

src/cljs/commiteth/ui_model.cljs
```diff
@@ -0,0 +1,121 @@
...
+
+    (= filter-type ::bounty-filter-type|value)
+    (str "$" (first filter-value) "-$" (second filter-value))
+
+    :else
+    (str filter-type " with val " filter-value)))
+
+(defn filter-bounties [filters-by-type bounties]
+  (let [filter-preds (->> filters-by-type
+                          (remove #(nil? (val %)))
```
siphiuel: Probably substitute with `(filter second)`.

src/cljs/commiteth/ui_model.cljs
```diff
@@ -0,0 +1,121 @@
...
+                          (map (fn [[filter-type filter-value]]
+                                 (let [filter-type-def    (bounty-filter-types-def filter-type)
+                                       pred               (::bounty-filter-type.predicate filter-type-def)
+                                       pre-pred-processor (::bounty-filter-type.pre-predicate-value-processor filter-type-def)
+                                       filter-value       (cond-> filter-value
+                                                                  pre-pred-processor (pre-pred-processor filter-value))]
```
siphiuel: Should probably be `pre-pred-processor pre-pred-processor`, explicitly passing `filter-value` in the form part of `cond->` is not necessary here.

siphiuel:
General notes:

1. I like the idea of namespaced keywords and type specifiers in the name.
2. However, long keyword names are a bit unwieldy. Can you please change formatting in `ui_model.cljs` so that lines are shorter? E.g., move map keys and values to separate lines.
3. Probably makes sense to move `commiteth.ui-model` to `commiteth.model.sort-filter` or something similar.
4. Code is nice and descriptive, but some doc strings on heavier fns and components would be nice.
5. Seems that `bounty-filter-remove.svg` and `icon-forward-white.svg` images are missing.

Thank you!

src/cljs/commiteth/subscriptions.cljs
pablodip: Added logic to `filter-bounties` to not do anything if no non-nil filters are passed, so that the fn is more powerful from inside.

src/cljs/commiteth/subscriptions.cljs
pablodip: I think that we should consider `:open-bounties` with no order semantic, that is why probably better to apply the sorting-type even if it is the default one.

src/cljs/commiteth/ui_model.cljs
pablodip: It is like that because `false` is a valid filter-value.

src/cljs/commiteth/ui_model.cljs
pablodip: Nice catch, fixed! :)

pablodip:
Thanks for reviewing! :)

Regarding the general notes:

> 1.) I like the idea of namespaced keywords and type specifiers in the name.

Nice! :))

> 2.) However, long keyword names are a bit unwieldy. Can you please change formatting in ui_model.cljs and bounties.cljs so that lines are shorter? E.g., move map keys and values and fn args to separate lines.

Agree that long keyword names are a bit cumbersome in length, although they are really useful in understandability by providing context and in refactoribility by providing uniqueness, and we normally have nice editors to autocomplete.

Just changed the longest lines by taking some code to vars and by using different lines in map keys and values.

And I wish the biggest problems we have with code are about the length of the lines! :))

> 3.) Probably makes sense to move commiteth.ui-model to commiteth.model.sort-filter or something similar.

I'm not totally convinced about `ui-model`, but I am less about `sort-filter`. The reasoning behind `ui-model` is not to say what it is in a general way, but to set a boundary to separate the domain of our web app from the rest of the things needed for that domain to work, like tools (re-frame constructs) and ui representation. That is why I placed all our domain logic there. By just saying `sort-filter` we could well place there related tooling things or even ui views related with sorting or filtering. I personally think that the whole SOB should get some re-structuring regarding domain oriented programming, but that would require an isolated discussion. My vote is to leave it as `ui-model` for now, where we can place more decoupled things related to our domain, but not anything else. Or maybe rename it to something like `app-domain` or something like that. What do you think?

> 4.) Code is nice and descriptive, but some doc strings on heavier fns and components would be nice.

Could you please indicate what functions are too heavier to be hard to understand just by reading the code? I will firstly try to make them smaller and more descriptive just by code, since I think it is far more useful and maintainable, and otherwise I will add some doc strings.

> 5.) Seems that bounty-filter-remove.svg and icon-forward-white.svg images are missing.

Just added, sorry! :))

> 6.) Filter probably should not stay enabled after navigating to a different page, e.g. Activities.

Changed to that behaviour! :)

pablodip:
Btw, also merged with develop!

pablodip:
Pending: add the `Claims` filter.

siphiuel:
On 3): we agree here, maybe I inadvertently caused some confusion. I was not suggesting to banish `ui-model` namespace altogether, but move it to a separate package `commiteth.model`, and there rename it to `sort-and-filter`. Namespaces for other domain entities would then be placed inside `commiteth.model` as well.

On 4): I'd say vote for predicate-related defns in `ui-model`: `bounty-filter-type-date-pre-predicate-value-processor`, `bounty-filter-values-by-type->predicates` to be a bit more documented.

pablodip:
3.) Totally agree that we should have a better structure for elements inside the domain. Just think that it requires some deep thinking and have in mind the whole system, including the backend, not just but the whole system. How about leaving `ui-model` for now and treat this topic deeply separately? I will probably create a swarm idea for this. :)

4.) Just added doc strings for those fns! Let me know if they are clear enough or if other fns need some doc string as well. :)

siphiuel:
3) Ok! Let's think about it together.
4) Thanks! These look good.

src/cljs/commiteth/ui_model.cljs
```diff
@@ -92,16 +102,22 @@
...
+                                                              boolean))}
+
+   ::bounty-filter-type|claims
+   {::bounty-filter-type.name      "Claims"
+    ::bounty-filter-type.category  ::bounty-filter-type-category|single-static-option
+    ::bounty-filter-type.options   bounty-filter-type-claims-options-def
+    ::bounty-filter-type.predicate (fn [filter-value bounty]
+                                     (condp = filter-value
+                                       ::bounty-filter-type-claims-option|no-claims
+                                       (<= 0 (:claim-count bounty))))}})
```
siphiuel: Should be `(= 0 (:claim-count bounty))`.

pablodip: Just changed! Actually I did that on purpose to make the code more resilient, so for example `(= 0 nil)` is false but `(<= 0 nil)` is true. But it is probably rather cryptic, so I just left the equality as you mentioned and we will see if we have any more needs.

siphiuel: Understand now, did not about that ClojureScript (or rather Javascript!) trick. In that case it should have been `>=`, not `<=`. But yes, it's rather cryptic, `(= 0 (or (:claim-count bounty) 0))` or `(-> bounty :claim-count some? not)` is longer but clearer. Clojure is good:)

Issue #177, opened 2017-12-04
Placeholder is displayed instead of new address in Ethereum address after you set new address  in "Manage payments" after you left and come back 


churik:
### Description
*Type*: Bug
*Summary*: 
When you set new address in  Ethereum address field and then you left tab "Manage payments", when you return you can see only placeholder is displayed, so user think that nothing is updated.

#### Expected behavior
you can see new address

#### Actual behavior
placeholder "0X0000000..." is displayed

### Reproduction

- Open Status
-  Set new address in "Manage payments" ) - I did it manually
-  Go to other tab - e.g. "bounties"
-  Go to "manage payments" again

Video: https://drive.google.com/open?id=1T6ZC69VtjObdMsyKzDnkFtY6sfGuELml

* Browser:  latest Chrome 62.0.3202.94

[comment]: # (OS version)
* Operating System: macOs high Sierra 10.13

Can be reproduced:
- when user set up wallet for first time manually
- when user update address manually

churik:
closed via https://github.com/status-im/open-bounty/pull/240

PR #240, opened 2018-01-29
[FIX #187] Fix address dropdown in My Payment Details

### LINKED ISSUE #187: Metamask address is not updated automatically in Payment Details
### Description

*Type*: Bug

*Summary*: 
when you setup Metamask, unlock it with your browser, login to SOB and go "My payment details" you see new address from MetaMask, but it is not really updated UNTIL you click update manually.
For you it looks like new address, but in database old address remains.

#### Expected behavior
address is updated in database (new address from MetaMask is set)

#### Actual behavior
address is updated on SOB but remains old (or empty, if it is first time) in DB
Only after you click UPDATE on SOB > Payment Details, it will be updated in DB

### Reproduction

- Open SOB
- Log in with user who set address manually or with new user
- Connect MetaMask for first time
- Go to "Payment Details"

siphiuel:
**Summary**

Following changes are included:

1. Do not preselect a Metamask address if it has not been set earlier by the user.
2. When Metamask is enabled, add a dummy 'Select address' option to the dropdown when no previous addresses were set.
3. Fix user address setting in 'My Payment Details`. Addresses were correctly stored in the DB, but not in re-frame's app-db.

**Status**: Finished.

### DIFF
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -10,18 +10,19 @@
      (merge props {:type      "text"
                    :value     @val-ratom
                    :on-change #(reset! val-ratom (-> % .-target .-value))})]))
-
 (defn dropdown [props title val-ratom items]
   (fn []
-    (if (= 1 (count items))
-      (reset! val-ratom (first items)))
     [:select.ui.basic.selection.dropdown
      (merge props {:on-change
-                   #(reset! val-ratom (-> % .-target .-value))})
-     (doall (for [item items]
-              ^{:key item} [:option
-                            {:value item}
-                            item]))]))
+                   #(let [selected-value (-> % .-target .-value)]
+                      (when (not= selected-value title)
+                        (reset! val-ratom selected-value)))
+                   :default-value (if (contains? (set items) @val-ratom) 
+                                    @val-ratom
+                                    title)})
+     (conj (doall (for [item items]
+                    ^{:key item} [:option {:value item} item]))
+           ^{:key title} [:option {:value title :disabled true} title])]))
 
 (defn moment-timestamp [time]
   (let [now (.now js/Date.)
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -13,23 +13,16 @@
 
 (defn dropdown [props title val-ratom items]
   "If val-ratom is set, preselect it in the dropdown.
-   Add value of val-ratom if it's missing from items list.
-   Otherwise, prepend title as a disabled option"
-  (let [items (cond->> items
-                  (and @val-ratom
-                       (not (contains? (set items) @val-ratom)))
-                  (into [@val-ratom])
-                  (not @val-ratom)
-                  (into [title]))]
-    (fn []
-      [:select.ui.basic.selection.dropdown
-       (merge props {:on-change
-                     #(reset! val-ratom (-> % .-target .-value))
-                     :default-value (or @val-ratom title)})
-       (for [item items]
-                ^{:key item} [:option {:value item
-                                       :disabled (= item title)} 
-                              item])])))
+   Otherwise, prepend title as a disabled option."
+  (fn []
+    [:select.ui.basic.selection.dropdown
+     (merge props {:on-change
+                   #(reset! val-ratom (-> % .-target .-value))
+                   :default-value (or @val-ratom title)})
+     (for [item items]
+       ^{:key item} [:option {:value item
+                              :disabled (= item title)} 
+                     item])]))
 
 (defn moment-timestamp [time]
   (let [now (.now js/Date.)
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -10,19 +10,26 @@
      (merge props {:type      "text"
                    :value     @val-ratom
                    :on-change #(reset! val-ratom (-> % .-target .-value))})]))
+
 (defn dropdown [props title val-ratom items]
-  (fn []
-    [:select.ui.basic.selection.dropdown
-     (merge props {:on-change
-                   #(let [selected-value (-> % .-target .-value)]
-                      (when (not= selected-value title)
-                        (reset! val-ratom selected-value)))
-                   :default-value (if (contains? (set items) @val-ratom) 
-                                    @val-ratom
-                                    title)})
-     (conj (doall (for [item items]
-                    ^{:key item} [:option {:value item} item]))
-           ^{:key title} [:option {:value title :disabled true} title])]))
+  "If val-ratom is set, preselect it in the dropdown.
+   Add value of val-ratom if it's missing from items list.
+   Otherwise, prepend title as a disabled option"
+  (let [items (cond-> items
+                  (and @val-ratom
+                       (not (contains? (set items) @val-ratom)))
+                  (conj @val-ratom)
+                  (not @val-ratom)
+                  (conj title))]
+    (fn []
+      [:select.ui.basic.selection.dropdown
+       (merge props {:on-change
+                     #(reset! val-ratom (-> % .-target .-value))
+                     :default-value (or @val-ratom title)})
+       (doall (for [item items]
+                ^{:key item} [:option {:value item
+                                       :disabled (= item title)} 
+                              item]))])))
 
 (defn moment-timestamp [time]
   (let [now (.now js/Date.)
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -15,21 +15,21 @@
   "If val-ratom is set, preselect it in the dropdown.
    Add value of val-ratom if it's missing from items list.
    Otherwise, prepend title as a disabled option"
-  (let [items (cond-> items
+  (let [items (cond->> items
                   (and @val-ratom
                        (not (contains? (set items) @val-ratom)))
-                  (conj @val-ratom)
+                  (into [@val-ratom])
                   (not @val-ratom)
-                  (conj title))]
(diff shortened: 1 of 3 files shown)


Chu127001:
I'm sorry, but I cannot review code itself. Only can test it when it will be deployed.

siphiuel:
Also fixes #177.

src/cljs/commiteth/common.cljs
```diff
@@ -12,16 +12,24 @@
...
-      (reset! val-ratom (first items)))
-    [:select.ui.basic.selection.dropdown
-     (merge props {:on-change
-                   #(reset! val-ratom (-> % .-target .-value))})
-     (doall (for [item items]
-              ^{:key item} [:option
-                            {:value item}
-                            item]))]))
+  "If val-ratom is set, preselect it in the dropdown.
+   Add value of val-ratom if it's missing from items list.
```
pablodip: Should we allow the value of the val-ratom to not be in the item coll? Wouldn't it be better to just require it, since the item coll tells us the order? Not sure if this requires changes in other places of the app though, in which case we shouldn't probably do it.

src/cljs/commiteth/common.cljs
```diff
@@ -12,16 +12,24 @@
...
+   Otherwise, prepend title as a disabled option"
+  (let [items (cond-> items
+                  (and @val-ratom
+                       (not (contains? (set items) @val-ratom)))
```
pablodip: Like the use of `contains?` with a set instead of just using the set as a function, since it is much clearer! :)

src/cljs/commiteth/common.cljs
```diff
@@ -12,16 +12,24 @@
...
+                  (conj @val-ratom)
+                  (not @val-ratom)
+                  (conj title))]
```
pablodip: Are we requiring the item coll to be a list? Since `conj` would behave differently with other types of colls. How about putting it the first element somehow even if it is not the most efficient way, so that at least we comply with semantics if other colls are used?

src/cljs/commiteth/common.cljs
```diff
@@ -12,16 +12,24 @@
...
+    (fn []
+      [:select.ui.basic.selection.dropdown
+       (merge props {:on-change
+                     #(reset! val-ratom (-> % .-target .-value))
+                     :default-value (or @val-ratom title)})
+       (doall (for [item items]
```
pablodip: Why is `doall` required? I think it is only required when dereferencing an atom inside of a lazy list, so that the atom gets the dereference correctly.

src/cljs/commiteth/common.cljs
siphiuel: This component is not used anywhere else, so things won't break if we change something here.

Why would we want to allow `val-ratom` not to be present in `items`? Because if e.g. MetaMask account is switched to one different from currently set in SOB, `items` will not contain it, hence user will not be able to select their current account in the dropdown.

Or do you suggest to properly populate `items` at the invocation point of `dropdown` instead?

src/cljs/commiteth/common.cljs
siphiuel: Right! Not required, removing.

src/cljs/commiteth/common.cljs
siphiuel: Agree. Using `into` instead.

siphiuel:
About `Select address` missing second time you enter `My Payment Details`: can you please describe exact steps? Failed to reproduce, used the following steps:

1. Set address to null in DB first (`update users set address=null where login='siphiuel'`)
2. Enable MetaMask.
3. Go to `My Payment Details`, two items present in the dropdown: `Select address` and MetaMask's account ID. Optionally can select second item in the drop-down, without clicking on `UPDATE`.
4. Go to `Bounties` tab, then back to `My Payment Details`. `Select address` is visible

src/cljs/commiteth/common.cljs
pablodip: Yeah, I meant that creating the whole collection of available items should probably be done outside, so that we don't need special behaviours/conditions inside, that is, just a coll of items and optionally one of those values as selected.

src/cljs/commiteth/common.cljs
```diff
@@ -12,16 +12,24 @@
...
-                            item]))]))
+  "If val-ratom is set, preselect it in the dropdown.
+   Add value of val-ratom if it's missing from items list.
+   Otherwise, prepend title as a disabled option"
+  (let [items (cond-> items
+                  (and @val-ratom
+                       (not (contains? (set items) @val-ratom)))
+                  (conj @val-ratom)
+                  (not @val-ratom)
+                  (into [title]))]
```
pablodip: Actually `into` has the same semantics as `conj`, since I think that it uses `conj` in the end.

By the way, instead of dealing with the semantics of clj collections, how about just generating the `option` html element before the `for` for the title when it is needed?

src/cljs/commiteth/update_address.cljs
```diff
@@ -3,14 +3,16 @@
...
             [cljs-web3.eth :as web3-eth]))
 
 
 (defn update-address-page []
   (let [db (rf/subscribe [:db])
         user (rf/subscribe [:user])
         updating-address (rf/subscribe [:get-in [:updating-address]])
-        address (r/atom @(rf/subscribe [:get-in [:user :address]]))]
+        address (r/atom (some-> @(rf/subscribe [:get-in [:user :address]]) 
+                                str/lower-case))]
```
pablodip: Why is this conversion to lower case really needed? I'm not an expert in Ethereum addresses, but I think that capitalisation plays a role for some semantics. Maybe @3esmit or @ can help us.

src/cljs/commiteth/update_address.cljs
```diff
@@ -3,14 +3,16 @@
...
     (fn []
       (let [web3 (:web3 @db)
             web3-accounts (when web3
```
pablodip: I know it already was like that, but how about taking the web3 accounts from a re-frame subscription?

src/cljs/commiteth/common.cljs
siphiuel: `into` will invoke `conj` on a vector `[title]`, so it will nicely put all seq items at the end, so no problem here.

src/cljs/commiteth/update_address.cljs
siphiuel: I ran into an issue with address string case. MetaMask account number contains mixed case symbols, for instance! E.g. this is one of my account numbers copied from MetaMask: `0x111d500fe567696D0224A7292D47AF11e8A4bCB4`.

So I think that this conversion is necessary when equality comparison is invoked. Of course we can force addresses into lowercase in the UI, and have a DB script to convert old possibly broken addresses that are already stored, but maybe this is a separate discussion.

pablodip: https://forum.ethereum.org/discussion/9220/eth-address-upper-and-lower-characters-does-not-matter

Just not sure what implications it has for the SOB right now.

Comparisons can probably be done in lowercase, but if possible it would be nice to save it to be able to perform that checksum in case we need it in the future.

src/cljs/commiteth/common.cljs
pablodip: Yeah, that treatment of collections is quite tricky in Clojure and is probably one of the things that it has to improve. :S

Btw, when possible could you please send a new comment instead of edit/update the previous one? To ensure that notifications are also sent for updates. Thanks! :)

siphiuel: Got you, ok! I thought notifications are delivered for comment updates as well.

src/cljs/commiteth/common.cljs
siphiuel: Moved items construction to invocation point. Dropdown preselects either the atom value or title.

src/cljs/commiteth/update_address.cljs
siphiuel: I suggest addressing it in scope of #242 .

siphiuel:
I wonder if this is not something related to async loading of user profile. I think I saw it once or twice during refreshing.

src/cljs/commiteth/update_address.cljs
siphiuel: Changed to comparisons in lower-case. Curiously, addresses returned by web3 js object are all lowercase.

pablodip:
There is some behaviour that doesn't convince me yet. When I have a different address in the database than the one that I have in MetaMask, I cannot just keep that previous address. Should that be correct? I think we should always allow to keep the previous information, and that it should be the default behaviour, no matter it is not in MetaMask, since it was already set before. cc @arash009 

Regarding that weird behaviour, it isn't just when the address is set, but when a different address is set, one that is not in MetaMask. Can you reproduce it like that?

siphiuel:
Yes you can keep that previous address. It will be displayed in the dropdown. I explicitly ensured that during implementation.

Regarding reproduction - yes I tried it with different addresses, but no success.

siphiuel:
In order to reduce confusion, a rundown of possible scenarios:

1. User does not have any address set at all.
  - MetaMask enabled: populate the dropdown with MM accounts, add a disabled `Select address` option as a first item in the dropdown
  - MetaMask disabled: display `0x000...` in the input field
2. User has an address set that is different from the one available in MM.
  - MetaMask enabled: populate dropdown with MM accounts, add user address as a first in the list of dropdown items
  - MetaMask disabled: display that address in an input
3. User has an address that that is included in the list of MM accounts.
  - MetaMask enabled: populate dropdown with MM accounts, select the one that is equal (case-insensitive comparison) to user's account stored in SOB.
  - MetaMask disabled: display that address in an input.

PR #270, opened 2018-02-09
makes the footer always at the bottom.

dashengSun:
Fix #268 

## Why do this changes:
* The height of the footer is 414px, shown in following attachment:
image
* Stick footer solution refer to [here](https://css-tricks.com/snippets/css/sticky-footer/)

PR #276, opened 2018-02-13
Footer is not pinned to the bottom fix

### LINKED ISSUE #268: Footer is not pinned to the bottom
### Description
*Type*: Bug
*Summary*: footer should be pinned to the page bottom, so I expect that no gap is shown between page bottom and footer.

#### Expected behavior
pinned footer

#### Actual behavior
a gap between footer and page bottom on some pages (which contains not a lot of content, i.e. one field)  

### Reproduction

- Open https://openbounty.status.im:444/app
- Log in
- Tap on Profile > My payment options

### Additional Information
* Browser: Chrome, Safari 
* Operating System: OSx
* Display: 15,4-inch (2880 x 1800)

It should work with the main operating systems (Windows, Linux and Mac) and the main browsers (Chrome, Firefox, IE and Safari)

CS76:
This commit fixes the footer gap on the bottom. 

Description: 
Semantic UI CSS defines a padding of 1em to the top and bottom of the segment element. Moving the footer outside the dev.ui.vertical.segment would avoid this gap in the bottom. Also, div.commiteth-footer has excessive margin on top (20px) changed this to 10px to maintain the flow of components (optional).

I also see the fix via margin adjustment (#270), this would not solve the issue as seen in the Manage payouts component. Also, it's not advisable to hard code margin based on the height of footer.

Fixes #268

### DIFF
--- a/src/cljs/commiteth/core.cljs
+++ b/src/cljs/commiteth/core.cljs
@@ -218,7 +218,7 @@
               [:h3.top-hunters-header "Top 5 hunters"]
               [:div.top-hunters-subheader "All time"]
               [top-hunters]]])]]
-        [footer]]])))
+        ][footer]])))
 
 (secretary/set-config! :prefix "#")
 
--- a/src/cljs/commiteth/core.cljs
+++ b/src/cljs/commiteth/core.cljs
@@ -217,8 +217,8 @@
              [:div.ui.container.top-hunters
               [:h3.top-hunters-header "Top 5 hunters"]
               [:div.top-hunters-subheader "All time"]
-              [top-hunters]]])]]
-        ][footer]])))
+              [top-hunters]]])]]]
+       [footer]])))
 
 (secretary/set-config! :prefix "#")
 
--- a/src/less/style.less
+++ b/src/less/style.less
@@ -984,7 +984,7 @@
     z-index: -1;
     display: flex;
     justify-content: center;
-    margin-top: 20px;
+    margin-top: 10px;
 }
 
 
--- a/src/less/style.less
+++ b/src/less/style.less
@@ -862,6 +862,8 @@
     box-shadow: none;
     border-radius: 0.3em;
 
+    min-height: calc(~"100vh - 400px");
+
     h3 {
         color: #474951;
     }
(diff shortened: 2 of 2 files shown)


pablodip:
@CS76 Thank you for contributing! :)

Just one concern: does this solution work for the operating systems and browsers mentioned in the issue?

CS76:
Will update the tests soon...

CS76:
Update:

> Removed the margin-top of the footer as it is no longer required.

> Also added min-height to the ui.vertical.element div which calculated based on the screen size dynamically (calc() - https://caniuse.com/#feat=calc )

Testing:
Tested it on the following platforms and so far seems to be working fine

=======
Mac

Google chrome
Version 63.0.3239.132 (Official Build) (64-bit)

Firefox
Version 58.0.2 (64-bit)

Safari
Version 11.0.3 (12604.5.6.1.1)

========
Windows

Edge
38.14393.0.0

IE11
11.576.14393.0

Firefox
48.0

=======
Linux (Ubuntu)

Firefox
Version 58.0.2 (64-bit)

Chrome
Version 61.0.3163.79 (Official Build) (64-bit)

pablodip:
@CS76 Thank you for those tests! We are going to test it in QA and will let you know. :)

churik:
Retested in:
- Mac OS 10.13.3 (Chrome 64, Safari 11, FF 58) - 2880 x 1800
- Windows 10 (IE 10, Chrome)

src/cljs/commiteth/core.cljs
```diff
@@ -218,7 +218,7 @@
               [:h3.top-hunters-header "Top 5 hunters"]
               [:div.top-hunters-subheader "All time"]
               [top-hunters]]])]]
-        [footer]]])))
+        ][footer]])))
```
martinklepsch: Is this intentional or accidental? If intentional the paren should be moved up a line before or after merging.

CS76: Yes, it's intentional. Shall I edit (push the parent above) and commit the changes here ??

CS76: fixed it now before the merge. Thanks for pointing it out.

Issue #269, opened 2018-02-09
When all filters are applied and filter 'Value' is set (from $1000), they are not aligned by height (FireFox, Chrome)


churik:
### Description
*Type*: Bug
*Summary*: want to see all filters nicely aligned by height regardless of what filters are currently applied. Now if all filters are applied, filter 'Value' filter continues contain one row while other filters contain 2 rows.

#### Expected behavior
like in "Safari" 

#### Actual behavior
<img width="794" alt="11" src="https://user-images.githubusercontent.com/4557972/36033335-69257f50-0db9-11e8-991a-cbfb35fd2321.png">

### Reproduction

 - open https://openbounty.status.im:444/app
 - select on 'Value' amount >1000 (i.e. from $1000 - 10000)
 - apply all other filters
 
### Additional Information
* Browser: Chrome, FireFox
* Operating System: macOS 10.13.1

PR #283, opened 2018-02-14
Fix vertical height in filter boxes

pablodip:
Fixes: #269 

Status: Finished


pablodip:
@martinklepsch We're not requiring GPG signature for now. Could you open a Github issue to start a discussion about whether we should require that?

pablodip:
@churik Is this PR testable without merging it to `develop`?

churik:
Tested in Chrome, Safari, FF (mac OS HighSierra)

PR #288, opened 2018-02-14
[Fix #259] Migrate to DB backed sessions

omartell:
Fixes: https://github.com/status-im/open-bounty/issues/259
Status: Finished

## Summary
This migrates from Immutant servlet based sessions to DB backed sessions using [ring.middlware.session](https://ring-clojure.github.io/ring/ring.middleware.session.html), [jdbc-ring-session](https://github.com/luminus-framework/jdbc-ring-session) and [ring-session-timeout](http://ring-clojure.github.io/ring-session-timeout/)

## Why?
- Need to potentially store more data in the session.
- Need to share session information across different application processes.
- A user's session should survive application restarts.

## Things to note:
- A new table `session_store` has been added to hold session information. This will be automatically created when the app starts.
- Any stale session records will be automatically deleted by a session cleaner that runs on the background. A session is considered stale after 6 hours of inactivity.

## QA Notes

(added by @martinklepsch)

- A successful login
- A successful logout
 - After 6 hours of inactivity, they should be logged out and redirected to the home page. The 6 hour period could be extracted to the config and maybe they can change it in their test environment to something shorter (10 min)?
- Session being still present after server restart would be an important thing to test here.

## Alternatives:
I think this might become a bottleneck in the future as every new session hits the DB, but not really sure about this as I don't know the current usage patterns. An alternative could be to use encrypted cookie based sessions, but this depends on the data that's goint to be stored in the session.

I'd appreciate your feedback on the approach as well as on the implementation.

More info:

- http://www.luminusweb.net/docs/sessions_cookies.html
- https://github.com/ring-clojure/ring/wiki/Sessions

PR #289, opened 2018-02-14
remove moment.js dependency

### LINKED ISSUE #278: Replace cljsjs.moment with goog.date helpers
### User Story

As a user, I want to OpenBounty to load and run fast so that I can get things done.

### Description

*Type*: TechDebt

*Summary*: Currently we depend on `cljsjs.moment` for a single function:

```clj
;; commiteth/common.cljs
(defn moment-timestamp [time]
  (let [now (.now js/Date.)
        js-time (clj->js time)]
    (.to (js/moment.utc) js-time)))
```

Moment.js adds ~20k (gzipped) to the build which impacts download and JS parsing time. Using `goog.date` we are able to achieve the same while also benefitting from advanced compilation strategies such as dead code elimination. I expect that with `goog.date` we might end up with this function not taking more than 1kb. 

[goog.i18n.DateTimeFormat](https://google.github.io/closure-library/api/goog.i18n.DateTimeFormat.html) might be all that's needed here.

rcullito:
and replace with goog.date.relative.

Fixes #278 

One aspect of this that I was unable to discern locally was what type the [time] parameter was previously? I assumed milliseconds, but am happy to add additional logic to the extent that it is stored in a different format.

### DIFF
--- a/src/cljs/commiteth/common.cljs
+++ b/src/cljs/commiteth/common.cljs
@@ -2,7 +2,7 @@
   (:require [reagent.core :as r]
             [re-frame.core :as rf]
             [clojure.string :as str]
-            [cljsjs.moment]))
+            [goog.date.relative]))
 
 (defn input [val-ratom props]
   (fn []
@@ -23,10 +23,10 @@
                             {:value item}
                             item]))]))
 
-(defn moment-timestamp [time]
-  (let [now (.now js/Date.)
-        js-time (clj->js time)]
-    (.to (js/moment.utc) js-time)))
+(defn relative-time [time]
+  "converts time in milliseconds to a relative form of '1 hour ago'"
+  (let [js-time (clj->js time)]
+    (goog.date.relative/format js-time)))
 
 (defn issue-url [owner repo number]
   (str "https://github.com/" owner "/" repo "/issues/" number))
--- a/project.clj
+++ b/project.clj
@@ -42,7 +42,6 @@
                  [cheshire "5.8.0"]
                  [mpg "1.3.0"]
                  [pandect "0.6.1"]
-                 [cljsjs/moment "2.17.1-1"]
                  [org.clojure/tools.nrepl "0.2.13"]
                  [com.cemerick/piggieback "0.2.2"]
                  [jarohen/chime "0.2.2"]
--- a/src/cljs/commiteth/activity.cljs
+++ b/src/cljs/commiteth/activity.cljs
@@ -1,7 +1,7 @@
 (ns commiteth.activity
   (:require [re-frame.core :as rf]
             [reagent.core :as r]
-            [commiteth.common :refer [moment-timestamp
+            [commiteth.common :refer [relative-time
                                       items-per-page
                                       display-data-page
                                       issue-url]]))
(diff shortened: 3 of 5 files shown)


src/cljs/commiteth/common.cljs
```diff
@@ -2,7 +2,7 @@
   (:require [reagent.core :as r]
             [re-frame.core :as rf]
             [clojure.string :as str]
-            [cljsjs.moment]))
+            [goog.date.relative :as relative]))
```
martinklepsch: When using a namespace only a few times I often prefer to not alias it, i.e.
```
[goog.date.relative]
```
This makes it easier to read the code without keeping all the aliases in your head.

This is general feedback, you don't need to change this to get this PR merged. 👍

rcullito: that feedback makes sense! just updated without the alias for the sake of consistency

rcullito:
Thank you for your feedback @martinklepsch! I love the project though did find some of the setup steps a bit tricky. 

Mainly I was flipping between the README and the riot chat for getting the geth node working (and am still not wholly confident in my process). Do you think this is strictly necessary to have in order to `lein run` or would there be value in designing a mock service so that this was not required to get the webapp itself working? 

I also wasn't wholly sure what role the github and Oauth apps are playing in this repo working locally, so more of an explanation around that might be helpful in order to come up with the right credentials? Thanks for bearing with all of my questions as I get up to speed!  Excited to keep contributing. 😺

src/cljs/commiteth/common.cljs
```diff
@@ -23,10 +23,10 @@
                             {:value item}
                             item]))]))
 
-(defn moment-timestamp [time]
-  (let [now (.now js/Date.)
-        js-time (clj->js time)]
-    (.to (js/moment.utc) js-time)))
+(defn relative-time [time]
```
pablodip: How about something like `time-millis->relative-str`?

pablodip:
@rcullito Could you please describe steps for QA to verify that sessions are working properly? You can do it in the description of the pull request.

martinklepsch:
@rcullito thanks for taking the time to write up your observations about the setup process:

> Mainly I was flipping between the README and the riot chat for getting the geth node working (and am still not wholly confident in my process). 

We're currently investigating using Infura to simplify this process. A mock service may also be an option but I believe Infura might be easier/quicker to implement for now.

> I also wasn't wholly sure what role the github and Oauth apps are playing in this repo working locally, so more of an explanation around that might be helpful in order to come up with the right credentials? 

Yeah that makes sense. It's a bit of a pain to set up all that stuff. I think for this specifically having some development authentication service which does not rely on Github could be very useful. I'll note that and see if we can make it happen.

> Thanks for bearing with all of my questions as I get up to speed! Excited to keep contributing. 😺

Awesome to have you on board! 🤗 

---


> @rcullito Could you please describe steps for QA to verify that sessions are working properly? You can do it in the description of the pull request.

@pablodip this PR has nothing to with sessions maybe you wanted to add that to #288? 

--- 

In the spirit of [optimistic merging](http://hintjens.com/blog:106) I will merge this PR. I didn't test it locally but according to the Moment.js / Closure docs the changes seem sound and my impression is that @rcullito did test this properly. If it's not working, I'll take the blame.

pablodip:
> @pablodip this PR has nothing to with sessions maybe you wanted to add that to #288?

Sorry, I forgot to modify the message, but was intended for this PR.

> In the spirit of optimistic merging I will merge this PR. I didn't test it locally but according to the Moment.js / Closure docs the changes seem sound and my impression is that @rcullito did test this properly. If it's not working, I'll take the blame.

Let's try to pass all PRs that modify code that is used by the system (not lein config for example) pass through QA, please, especially because we almost don't have any automated tests and don't use CI to validate.

churik:
Tested in #311

Issue #271, opened 2018-02-11
Production Workflow


pablodip:
I am feeling two problems in the current production workflow:

1.) We have several places to organise production: right now two projects on Github. Besides, some issues don't even fit in those two projects, so can't follow any workflow and their progress is more difficult. All that makes us not know well where to look to know what is going on, and avoid us to visualise well the status of the whole production.

2.) The columns in the Github projects don't show well the status of the issues. For example some of the issues in "TODO" aren't actually ready to "do", implement, but many of them are to define, or to design. That makes us not visualise well where the blocking part in the production is.

My proposals to solve those problems are these:

1.) Have just one project on Github, with the normal workflow that we want to follow in the SOB. We can put there any issues that needs to follow the normal workflow, which are almost all issues. And if we ever have any other workflow, we can have a new project, but only for a different workflow, not for a different group of cards. If we want to group cards by type, we can just tag them, and they can be filtered in the project.

2.) For this we decided to use tags, but we just defined two "designing" and "developing", and thinking more about it I don't think that tags makes us visualse well the status. That is why visualising the status is precisely the purpose of columns. So even though we might end up with more columns, if we really have all those steps in the production, I think it is worth using them. I would add these ones, "To Define", "Defining" "To Design", "Designing", "To Develop", "Developing". Those are just four more columns than the current ones.

I think that both of those changes together will allow us to visualise better the whole production and to make it flow more. :)

Let me know what you think. :)

arash009:
Thanks for the suggestions @pablodip . I'm happy with the point 2 re additional columns. On point 1 and having just 1 project, I think its a good idea to combine the 2 current projects, but I'm unsure about putting the entire backlog in the project then.

For one, we have alot of issues and we continuously create more. Having all new items in the project means having to continuously priorities the to-do list (which is both good and bad)

The other point is about having a pre-defined scope of work for releases. If we can group certain features together, then we can size them, estimate them, and come up with a timeline for implementation and release. So once we deliver SOB v2, we can think about v3 and so on.

What if we have projects defined based on scope and timeframes? So any new urgent issues that come up can be added to current project, but then we can have prioritisation and scoping exercise to define scope for SOB v3 (which could have say a Q2 release date), and then others?

pablodip:
@arash009 Those are nice problems to solve. :) I am going to number them so that we can distinguish them better.

[...]

Issue #221, opened 2018-01-17
Define code review workflow


pablodip:
### Description

I think it would be nice to make explicit a code review workflow, since now there are more people working on the SOB, and we also probably aim to set an example to work better with bounties.

### User Stories

As a developer I want to have other devs review my code so that the code that goes to production is checked by other people and they can spot improvements or mistakes I didn't.

As a developer I want to be able to review commits before they go to production, so that I want all changes be made through pull requests.

As a developer I want that an automatic tool helps me follow the code review workflow so that I can get used to them and follow them better.

As a new developer/external contributor I want to read about the code review workflow in the docs so that I know how things work here and start contributing better.

As a product manager I want devs to review their code to each other so that better code is done.

### Expected Behaviour

Have a defined code review workflow and be able to read it somewhere, could be in the README.

### Actual Behaviour

No defined workflow.

### Proposed workflow

* Make all changes through pull requests against QA (develop).

* If there is some urgent bug that needs to be fixed in production (master), make a pull request against develop if it is not polluted with other changes, or against master directly otherwise. If so, make a pull request to merge master with develop so that develop is keep up to date when it is bypassed.

* Have at least two request approvals in pull requests before merging them. There are going to be four devs now (including @dmitryn that currently is more with status-react), so I think it is doable.

* If there is a pull request to fix something very urgent in production and there aren't other devs online to review it, merge it directly.

* Enable "Require pull request reviews before merging" and "Dismiss stale pull request approvals when new commits are pushed" on Github: https://help.github.com/articles/about-required-reviews-for-pull-requests/

What do you think @arash009  @dmitryn @tpatja  @siphiuel ? :)

arash009:
I'm happy with the defined approach. My only comment would be on point 3 to have two approvals before a merge. This might slow the pace of work. Maybe one approval is enough for simple changes, and two approvals for large complex implementations.

@oskarth if you have any learnings/lessons learnt from processes by other swarms.

pablodip:
@arash009 On one hand I agree, since that could move pull requests faster into production. On the other I don't agree, since by requiring two request approvals we ensure that the knowledge and changes are better distributed in the team. Let's see anyway if @oskarth says something about this.

One more thing I just though of that could be defined in the workflow.

Who merges the PRs, the creator or the reviewers? I think that if the creator is internal, then the creator. Otherwise the reviewer, since the creator won't have permissions. This is because the creator is the one responsible to make it arrive to production.

[...]

Issue #212, opened 2018-01-15
Determine git workflow


pablodip:
### Description

Currently master has more commits than develop, which means than changes have gone to master without passing through develop, and that develop isn't up to date with master.

Those two things have probably happened because of the complexity of maintaining both branches (explained later). The point is that those two things complicate collaboration and stability, because other collaborators can't review what is directly committed into master (production).

Having both master and develop makes sense when working with releases. But that workflow comes with a cost: maintenance complexity and slower flow of changes into production.

The develop branch is just a buffer to group changes, which are moved together into master. It represents the workflow of a release: it groups changes, and those changes are released from time to time.

Two more examples of things that are more difficult with both master and develop:

* Priority in changes: if you have some changes buffered in develop and you want to put something into production before those buffered changes, you need to either change directly master (which is described above) or merge the buffered changes with the new change (and the buffered changes might not be well tested yet, so this is bad for stability).
* Agility in the decision of changes: if you have some changes buffered in develop and you want to delay one, or to make only one, you need to separate that change from develop, since it is mixed with other changes.

The SOB doesn't have releases, but it is just a production environment (master). So I think that we are paying a cost that we don't need to.

I think the SOB fits better for a continuous delivery environment, that is, once things are ready and tested, they go directly to production. That way we wouldn't have all issues described above.

### User Stories

As a developer I want to be able to review commits before they go to production, so that I want all changes be made through pull requests.

As a developer I want to have as less workflow complexity as possible, so that I want pull requests be merged directly into master.

As a product manager I want to have a simple and powerful flow to production so that I can work better with the needs of the product.

### Proposed Changes

* Protect the master branch on Github: https://help.github.com/articles/configuring-protected-branches/

* Make pull requests pointing to master directly, and stop using the develop branch.

* Make master the default branch on Github: https://help.github.com/articles/setting-the-default-branch/.

tpatja:
I disagree with the proposition of just using `master`. Like the README says, currently we have two continuously deployed environments - `develop` is auto-deployed to the shared test environment where we do QA and only when QA passes we merge to `master` after which the changes end up in production. The problem with just using `master` is that even if we review the changes, there is no easy way to test them. The shared test environment solves this, QA team currently does manual testing there and is working on automated tests.

`develop` is the default branch, because were getting PR's based against master. Also, `master` is already a protected branch.

pablodip:
@tpatja 

[...]

PR #290, opened 2018-02-15
Doc to define development workflow

churik:
First attempt to define workflow.
Based on #271, #221.
Should be extended with examples and GIT workflow (is still discussed  in #212), and info about bounty issues and contributing process (like described in https://wiki.status.im/Developer_Documentation#Contributing)


doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
+# Development process in Status Open Bounty
+
+We have a continuously deployed version tracking the `develop` branch live at [test environment](https://openbounty.status.im:444). It uses the [Ropsten](https://ropsten.io/) Ethereum testnet. Any one is welcome to use it and contribute to Open Bounty.
+ 
+Any help is greatly appreciated!
+Currently we use two projects -  `Pipeline For Issues` and `Pipeline For Pull Requests`, issues and pull requests in our repository are passing through all or several stages described below.
+
+## Pipeline For Issues
+
+Issues are added to this project by product owner.
```
pablodip: Why only by the product owner? I think that the main new features are added by the product owner, but anyone can help adding bugs there, and the ones about technicalities are probably added by the tech people.

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### To define
+This is backlog for all features/issues/enhancements which we want to include to development process. 
+
+All issues here should be marked with:
+* type - `bug`, `tech-debt`,` enhancement`, `proposal` labels
```
pablodip: What is the advantage of having `tech-debt` issues in this workflow? The only one I can think of is prioritization, but the rest of the workflow (definition and design) doesn't help much in technical debt.

Also: how about having proposals outside the workflows, and only once they evolve from proposal to wanted issue they get this workflow? Issues in `To Define` should probably be more than a proposal, and `To Define` is the first step of this workflow.

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+* priority - `Prio: high`, `Prio: med`, `Prio: low` labels
```
pablodip: Should we ignore then order of columns and only use these tags for prioritization?

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### In Bounty
```
pablodip: I think that it would be better to position it at the right of `To Develop`, since a bounty issue is already defined, designed, and ready for development, but we just decide to put it as a bounty at that point.

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+The store with issues which are open bounties. When we put funds to issue and it shows up in [status open bounty](https://openbounty.status.im), we move issue here.
+
+#### Defining
+The column is intended for issues not completely clear or for features, that should be splitted to smaller issues in order to go ahead.
+After defining all issues that are intended to develop should have size label(`Size: S` - 2-3 hours, `Size: M` - 6-8 hours, `Size: L` - 8-24 hours)
```
pablodip: XL? It also exists! :))

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+#### Developing
+This is for issues that are currently developing, so pull requests assosiated with issues have to be placed in `Pipeline for Pull Request` project. 
+#### Done
+It stores issues with merged to `master` pull requests.
+## Pipeline For Pull Requests
+
+#### Developing
+Contains all open pull requests (hereinafter PRs) that already assosiated with issues.
+#### Contributor
+It keeps all PRs from external contributors which should pass `Reviewing` stage.
```
pablodip: Not sure to get the point of this column. What is the difference with `Developing`?

I think that we should try to treat as much as possible internal and external work, unless we have a different specific workflow, but reviewing is something required for both internal and external pull requests.

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+#### Reviewing
+The storage for all  PRs that pass reviewing process.
+Review  process is discussed [here](https://github.com/status-im/open-bounty/issues/221)
+Main points:
+* PR is quick fix (from core contributor) - 1 approval from core contributor
```
pablodip: I would say `trivial` instead of quick fix, since we always aim for 2.

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+* PR is complex (or from external contributor) - 2 or more approvals from core contributors
```
pablodip: I would say `normal`, so that we tread 2 reviews as a normal thing. :)

doc/development_workflow.md
pablodip: Okay, I get it now from something from below. How about calling it`Reviewing, waiting for contributor`? And we would also have `Reviewing`, the normal one.

doc/development_workflow.md
```diff
@@ -0,0 +1,60 @@
...
+* PR need to be based on and opened against the `develop` branch.
+* If a PR has undergone review and requires changes from author, move it back to `Contributor` column
+#### To test
```
pablodip: How about having a `To review` column as well? So that prs batch there waiting for review.

pablodip:
I think that it would be nice to add as well that the whole team is responsible to keep the projects with accurate information, so that we should all have a look at things from time to time and move anything that is outdated or wrongly placed.

doc/development_workflow.md
martinklepsch: `Blocked` may also be a more generally usable label that we could add to PRs waiting for adjustments.

doc/development_workflow.md
martinklepsch: I assume order may be relevant if there are issues with same priority?

doc/development_workflow.md
martinklepsch: I would suggest to avoid specific formulations like that and maybe phrase it more generally:

> The number of reviewers should be proportional to the complexity of the change.

This gives us some freedom to do the right thing in specific situations without being tied to too specific rules.

martinklepsch:
One question that I think is still a bit unanswered is: Do all issues go through this process and do we only work on issues that are part of this project or do we also work on other stuff outside this project? 

I think having everything go through the project makes most sense.

doc/development_workflow.md
churik: agree, will fix

doc/development_workflow.md
churik: Let it be like this:
`Prio: high`, `Prio: med`, `Prio: low` labels, inside issues with same priority - higher priority in the top. I think we always can reorder issues in our meeting standups or PO can move it to the top if it is crucial for some reason.

doc/development_workflow.md
churik: Hm, may be you are right. I just wondering - we have a lot of redundant issues in `To develop`  column which are intended mostly for external contributors. If all developers are happy to have all these issues in `To develop` - I don't see any problem to remove the column

doc/development_workflow.md
churik: xl
Not in current repo, I think) but we can add it.
`Size: XL` - 24-40 hours - is OK? :)

doc/development_workflow.md
churik: @pablodip as you wish, let's call it `Reviewing, waiting for contributor`

@martinklepsch agree with `Blocked`, will add info

doc/development_workflow.md
churik: np :)

doc/development_workflow.md
churik: thanks, good suggestions

doc/development_workflow.md
churik: Makes sense, will add

churik:
@pablodip yes, it is supposed to be but documents matter, so I will add it.

@martinklepsch  having everything go through the project makes most sense, absolutely.
The only thing that all discussions like  #271 or `Proposals` - which are not converted to enhancements live only in GH Repo but aren't added in the projects. So everything that need to be developed OR tested should go through project (or projects)

doc/development_workflow.md
pablodip: @martinklepsch 

> The number of reviewers should be proportional to the complexity of the change.

To me this is a suggestion, not a process, and totally open to interpretation. :)) Actually it is like an oxymoron, since `should` is a requirement, but here it is a requirement totally open to interpretation, which cancels the requirement. :)

Agree with not having a strict process, and to be able to respond to urgent things if needed, but I think that we need to be more specific so that we can have some expectations.

doc/development_workflow.md
pablodip: > So, I don't see the reason to exclude these issues from Pipeline for issues

What I try to find is a reason to include it, not exclude it. :))

But totally okay with having them.

And of course the associated pull requests will have to go through `Pipeline for issues`, I just meant to the issues. :)

doc/development_workflow.md
pablodip: Oh, I didn't mean to remove the column `In Bounty`, but just to move it at the right of `To Develop`.

doc/development_workflow.md
pablodip: Can't remember. :S cc @arash009 @andytudhope

doc/development_workflow.md
pablodip: Should we use a label or a column for `Blocked`? I like more a column so that we can visualise them better in the flow, and because when something is blocked it isn't being neither developed nor reviewed.

martinklepsch: Fine with either. :)

doc/development_workflow.md
martinklepsch: To me this document is more like a guideline and not so much a specification. So not sure if I agree that `should` should be interpreted as defining a requirement here. IMHO we should aim to give advice around review requirements instead of defining strict rules.

pablodip: Why exactly is it too strict for you to aim for 2 approvals in normal cases? I personally don't perceive it as a strict rule, but as a healthy custom that helps checking code and spreading knowledge.

doc/development_workflow.md
churik: Ok, agree

doc/development_workflow.md
churik: As for me ,'Blocked' label is more convenient jut because it can be applied fir issues and for PRs. Two columns in 2 pipelines for me looks a bit redundant
@pablodip ?

pablodip: To me reusability shouldn't guide whether label or column, but actual stages in the flow of the pipeline. Anyway, let's go for label if you like it more and we can see how it works once we use it. :)

doc/development_workflow.md
```diff
@@ -0,0 +1,71 @@
...
+This is backlog for all features/issues/enhancements which we want to include to development process. 
+
+All issues here should be marked with:
+* type - `bug`, `tech-debt`,` enhancement` labels. Issues with `proposal` label should be converted to `bug`, `tech-debt`,` enhancement` before addidng to project.
+* priority - `Prio: high`, `Prio: med`, `Prio: low` labels. On the board inside issues with same priority sorting from higher to lower priority is applied.
+
+#### Defining
+The column is intended for issues not completely clear or for features, that should be splitted to smaller issues in order to go ahead.
+After defining all issues that are intended to develop should have size label
+*`Size: XS` - 1-2 hours,
```
martinklepsch: line break missing here, list formatting is broken

PR #279, opened 2018-02-13
Currency filter should use OR

pablodip:
Fixes #275 

Status: Finished

PR #298, opened 2018-02-16
Fix filter currency eth

### LINKED ISSUE #275: Filter by currency use 'AND' instead of 'OR'
### Description
As it turned out, most of users expect that filter by currency will use 'AND' instead of 'OR'

#### Expected behavior
'ETH' and 'SNT' enabled return all bounties with 'ETH', 'SNT' or with both of them

#### Actual behavior
'ETH' and 'SNT' enabled return only bounties with both 'ETH' and 'SNT'
ethsnt

### Reproduction

- Open Status
- tap on 'Currency'
- Tick 'ETH' and 'SNT' 

### Additional Information

* Browser: Chrome
* Operating System: Mac OS

pablodip:
Fixes #275 

Status: Finished.

Related: #279

### DIFF
--- a/src/cljs/commiteth/ui_model.cljs
+++ b/src/cljs/commiteth/ui_model.cljs
@@ -87,8 +87,8 @@
     ::bounty-filter-type.category                      ::bounty-filter-type-category|multiple-dynamic-options
     ::bounty-filter-type.re-frame-subs-key-for-options :commiteth.subscriptions/open-bounties-currencies
     ::bounty-filter-type.predicate                     (fn [filter-value bounty]
-                                                         (or (and (contains? #{"ETH"} filter-value)
-                                                                  (< 0 (:balance-eth bounty)))
+                                                         (or (and (contains? filter-value "ETH")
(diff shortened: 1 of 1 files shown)


siphiuel:
Just noticed that `:balance-eth` bounty field is forcibly converted to string via `(update :balance-eth eth-decimal->str)` in `commiteth.routes.services/prettify-bounty-items`. So it arrives to CLJS as a string already. Maybe a better idea would be to convert them to string in the subscription that bounty page is consuming, not right away in `open-bounties`.

pablodip:
@siphiuel How about doing that general data fixing in a bigger refactoring so that we ensure that this change doesn't affect any existent behaviour? :)

siphiuel:
Yes, agree it might hurt down the road.

churik:
### Environments:
* macOS HighSierra 10.13, Chrome  64
* macOS HighSierra 10.13, Safari  11
* macOS HighSierra 10.13, FireFox  58
### Tested functionality: 
verified that issue is not reproducible
### Regression: 
checked combination with 'Value', 'Owner' filter

PR #311, opened 2018-02-19
fix relative time display for long-ago dates

martinklepsch:
Fix for a problem reported by @churik where relative dates were not shown.

This happens with dates older than 2 weeks. I changed the code so that a human-friendly date is displayed instead of a relative statement.

### DIFF
--- a/src/cljs/commiteth/activity.cljs
+++ b/src/cljs/commiteth/activity.cljs
@@ -1,7 +1,7 @@
 (ns commiteth.activity
   (:require [re-frame.core :as rf]
             [reagent.core :as r]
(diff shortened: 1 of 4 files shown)


churik:
fix for #289 which is currently merged in `develop`.
Should be retested in develop

churik:
### Environments:
* macOS HighSierra 10.13, Chrome  64
* macOS HighSierra 10.13, Safari  11
* macOS HighSierra 10.13, FireFox  58
### Tested functionality: 
verified that old dates (>2 weeks are displaying in human-readable format, like `January 23`) in Open Bounties, Activity feed
### Regression: 
checked filter by date

Issue #223, opened 2018-01-18
Script to create a private Ethereum network for dev and test environments


pablodip:
### Description

Isolated dev and test environments help create better software, since those isolated environments are not polluted with external states (they are deterministic) and they are normally very fast, and local.

Current dev and test environments of the SOB depend on the public test network, which make the environments not being deterministic, nor fast (it takes a good time to set up), nor local.

I think this could be solved by having a script in the repo to create a private Ethereum network so that network can be used most of the times when developing and when running tests.

### User Stories

As a developer/CI server I want to have a script to run to create a private Ethereum network, so that I can have an isolated, deterministic, fast and local environment.

### Expected Behaviour

Have a script in the repo, probably in the `scripts` directory, that I can run and create a private Ethereum network.

### Actual Behaviour

Work with the public test Ethereum network.

### Some notes

* The script should allow to create several networks if called in parallel with different parameters.
* The script should create a network with a few accounts that have some ethers.
* The network should have the contracts used by the SOB.
* The network should have a few tokens deployed so that the SOB can use them.
* The network should be created in a deterministic way if possible, that is, providing always an equal network, with the same accounts, and the same contracts, and the same addresses.
* The network should be done with [Ganache](https://github.com/trufflesuite/ganache) or something similar that helps debugging.

3esmit:
I agree to have the deploy of contracts automated, user just need run the RPC node willing to deploy them, but the script that would run it is Truffle (and possibly Ganashe) 

As @jeluard mentioned, Ganache is an out-of-the-box solution, that works with truffle. Truffle is an awesome development framework, it contains testcases and deploy (migration) scripts.

As we are using a "contract library" (mastercopy) with forwarders created by a factory contract, the probable path to follow is implement a migrations to deploy the "contract library" (mastercopy) and the factory.
Some changes would be needed in the contract to make this auto-deploy possible, because the current contract being used have the "contract library" (mastercopy) hardcoded in a Stub contract that is deployed by a factory. You can read this in line 222 of factory contract. https://etherscan.io/address/0x3b9a3c062bdb640b5039c0ccda4157737d732f95#code

I was working in changes that would make this auto deploy possible, it can be found here https://github.com/status-im/open-bounty/tree/a3931724123edb938d6fc7dd7fd42cace5f3ece8/contracts
This was reverted because of an incompatibility with newer versions of solidity and web3j used by SOB-bot to understand ABI and other ethereum stuff. 
[...]

PR #313, opened 2018-02-20
Feature/seed postgres script

### LINKED ISSUE #312: seed postgres with sample activity data
### User Story

As a developer, when starting Open Bounty locally it would be nice to see example bounties and activity in the dashboard. 

A great deal of the core functionality within the repo is, at heart, about displaying this data in an easily understandable way. In addition to improving the developer experience, this would also be a good sanity check for examining different use cases of local changes, and spotting bugs before they get to a later stage in the QA process. 

### Description

*Type*: Feature

*Summary*:
This could take the form of a migration script that could be run simply with `lein seed` or some variant: a one off job that that is not necessarily required by the build process, but exists in the case that it offers assistance and prevents developers from reimplementing the same functionality.

rcullito:
Fixes #312 

This is very much still a work in progress, but is meant as the beginning of a seed script in order to populate postgres with human readable values to aid in local development. 

I wanted to start with users and make the PR public to the extent others might offer feedback on direction, etc... but the plan is to make a similar scaffold for issues, pull requests, and other entities in order to construct meaningful `activity` and `bounties` in the dashboard. 

Eventually this logic will live as a lein or boot task, but for now with just repl testing it was easiest to get started by taking advantage of the existing functions in the commiteth.db namespaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,16 @@ Landing page is static and different CSS and JS due to time constraints.
 
 This copies over necessary artifacts to `resources` dir.
 
+### Seed postgres for local development
+
+Postgres can be seeded with initial values to make sure that your local SOB app loads with some data. Just
+include the number of entities you wish to create as the first argument to `seed-postgres`, such as:
+
+```
+lein run -m commiteth.db.seed/seed-postgres 3
+```
+
+*Note, it may take a minute for all of your newly created entities to become visible in the local app.
 
 ## More info
 Detailed information on code structure, troubleshooting, etc. can be found [here](doc/README.md).
--- a/project.clj
+++ b/project.clj
@@ -50,7 +50,10 @@
                  [akiroz.re-frame/storage "0.1.2"]
                  [cljsjs/chartjs "2.6.0-0"]
                  [org.web3j/core "2.3.1"]
-                 [cljs-web3 "0.19.0-0-7"]]
+                 [cljs-web3 "0.19.0-0-7"]
+                 [reifyhealth/specmonstah "1.0.0"]
+                 [clojure-future-spec "1.9.0-beta4"]
+                 [org.clojure/test.check "0.9.0"]]
 
   :min-lein-version "2.0.0"
   :source-paths ["src/clj" "src/cljc"]
--- a/src/clj/commiteth/db/seed.clj
+++ b/src/clj/commiteth/db/seed.clj
@@ -1,0 +1,191 @@
+(ns commiteth.db.seed
+  (:require [clojure.spec.alpha :as s]
+            [clojure.spec.gen.alpha :as gen]
+            [clojure.future :refer :all]
+            [reifyhealth.specmonstah.core :as rs]
+            [commiteth.config :refer [env]]
+            [commiteth.db.core :refer [*db*] :as db]
+            [mount.core :as mount]
+            [clojure.java.jdbc :as j]))
+
+(def non-nil-pos-int
+  (s/and pos-int? (complement nil?)))
+
+;; names taken from https://www.ssa.gov/oact/babynames/decades/century.html
+(def female-names
+  '("mary" "patricia" "jennifer" "elizabeth" "linda" "barbara" "susan" "jessica" "margaret"))
+
+(def male-names
+  '("john" "robert" "william" "david" "richard" "joseph" "thomas" "charles"))
+
+(def email-regex #"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,63}$")
+
+(def base-address-generator (gen/fmap #(apply str %)
+           (gen/vector (gen/char-alphanumeric) 40)))
+
+(s/def ::id non-nil-pos-int)
+
+(s/def ::name (s/with-gen string?
+                    #(s/gen (-> male-names
+                                (into female-names)
+                                set))))
+
+(s/def ::login (s/and string? #(< (count %) 8)))
+
+(s/def ::email (s/with-gen (s/and string? #(re-matches email-regex %))
+                 #(gen/fmap (fn [username] (str username "@status.im"))
+                            (gen/string-alphanumeric))))
+
+(s/def ::avatar_url (s/with-gen string?
+                      #(gen/fmap (fn [username] (str "https://api.adorable.io/avatars/285/" username "@adorable.png"))
+                                 (s/gen ::name))))
+#_
+(s/def ::address (s/with-gen (s/and string? #(= 42 (count %)))
+                   #(gen/fmap (fn [rest-of-address] (str "0x" rest-of-address))
+                              base-address-generator)))
+
+(s/def ::address (s/with-gen (s/and string? #(= 42 (count %)))
+                   ;; this address needs to be valid for ropsten
+                   #(s/gen #{"0xFC5A9d404edeDb10AA6fd379b07324aAA76019F9"})))
+
+(s/def ::user (s/keys :req-un [::id
+                               ::name
+                               ::login
+                               ::email
+                               ::avatar_url
+                               ::address]))
+
+(s/def ::repo_id non-nil-pos-int)
+(s/def ::user_id non-nil-pos-int)
+(s/def ::owner string?)
+
+(def bounty-repos
+  #{"status-react" "open-bounty" "status-go" "ETHDenver"})
+
+(s/def ::repo (s/with-gen string?
+                #(s/gen bounty-repos)))
+
+(s/def ::hook_id pos-int?)
+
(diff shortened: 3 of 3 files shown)


pablodip:
We're in the process of using spec (https://github.com/status-im/ideas/issues/77), but I'm totally okay with having something to seed the database like this for now to help work with the UI.

rcullito:
@pablodip thanks for offering some feedback. That's a good point about spec. I enjoyed reading your comments on the other thread relating to it. 

I've already pulled in the `test.check` library here, so would be totally open to creating the seed data via spec generators. What I didn't want to do was intertwine the one off nature of a seed script with (what I view as the much larger) task of defining a single source of truth for the data within the application. 

But it seems easy to conceive of a world where scripts such as these pull in the spec from another module and use it as a blueprint for data generation? Does that align with your vision on where things area headed?

rcullito:
@pablodip will the decision to move to spec be accompanied by a clojure version bump from 1.8 to 1.9?

pablodip:
> But it seems easy to conceive of a world where scripts such as these pull in the spec from another module and use it as a blueprint for data generation? Does that align with your vision on where things area headed?

To me each module should be responsible of its own seeding, being able to use operations and some data specifications from other modules. But since we don't have modules for now, I am totally okay with having this in a single namespace as you are facing it. :)

Regarding `spec` or `test.check`, feel free to use the one that helps you more to describe the problem.

> will the decision to move to spec be accompanied by a clojure version bump from 1.8 to 1.9?

Yep, but probably not as part of this issue. If you feel like it, feel free to open a pull request with just that change so that we can test it separately.

By the way, as for the relations of data (a tricky part), you might find useful this library: https://github.com/reifyhealth/specmonstah It requires Clojure 1.9 though I think.

rcullito:
@pablodip thanks for the feedback, that all makes sense. Already it feels like the seed namespace would grow unwieldy if extended to all entities, so I will keep an eye out for how to tease it apart. Specmonstah seems worth evaluating too. Do you have any equivalent utilities that you like for setting the state of the ethereum network?

rcullito:
@pablodip the 1.9.0 PR landed here: https://github.com/status-im/open-bounty/pull/319

As I mentioned in the comments, if there are other steps you would be specifically looking for with a change of this kind - beyond just passing CI - let me know and I can try and validate those.

pablodip:
> Already it feels like the seed namespace would grow unwieldy if extended to all entities, so I will keep an eye out for how to tease it apart.

Feel free to separate the seeding code in different files if you feel more comfortable with that, just try to keep it under the same root if possible, for example all under `commiteth.db.seed`, so ``commiteth.db.seed.user`, etc.

> Specmonstah seems worth evaluating too. Do you have any equivalent utilities that you like for setting the state of the ethereum network?

We have an issue to have a private network so that we can have a more deterministic state in that sense: https://github.com/status-im/open-bounty/issues/223 But the issue is still pending.

Apart from that, since the SOB is right now based on GitHub notifications (which I personally don't like), my way of proceeding for a correct whole state would probably be to create some data in Github through the API, and let that flow in the rest of the system. But as I said, I am totally okay with just seeding the database for now and as a first step, since that already helps to work with the web app.

rcullito:
@pablodip thanks for your feedback on this. I've since seen how much integration with github matters for effective testing of the open bounty repo, and how via the webhooks various entities are inserted into postgres during the course of normal application use. 

Having said that, do you think there is value in still maintaining this PR or merging it in? I find the spec definitions and the specmonstah schema relations to be helpful declarations... but in terms of generating data, any default generators seem to be insufficient. One thought I had was to leave this responsibility up to the individual user. That way some helpful guard rails are in place but this wouldn't presume to seed valid data, that responsibility is still on the consumer to use their own credible repo, user, and issue ids? I wanted to check with you on whether this seemed sensible for proceeding further and hopefully wrapping this up.

martinklepsch:
@rcullito I think this kind of database seeding is still useful despite the fact that there are more challenges related to "bootstrapping" a development environment. So yes, lets push forward with this please.

My hope is that we can model our use of GitHub as a component which can then be swapped (mocked) during testing or local development. This should be handled separately though — maybe as part of #327. 

All that said it will probably take a week more or so to get this merged as we're currently dealing with some infrastructure problems that prevent us from properly testing etc.

rcullito:
@martinklepsch thanks for checking in on this! (and welcome back).

The Github mocking service totally seems useful, so that's great to hear ideas around it are already brewing. In light of that, I guess I'm just trying to figure out what more this current piece should tackle beyond linking proper repos, users, issues, and PRs? 

It seemed to me after my conversations with @siphiuel  that the next step for generating meaningful state was the creation of the `bounty` label on the issue itself? Do you think the transaction data downstream of that event should be seeded via this postgres script or is that the responsibility of the GH mocking service? (I'm still a little fuzzy on this contract data because that's about as far as I've made it in my own explorations lol).

rcullito:
Revisiting this today. Exit criteria include:

- [ ] update fields and foreign keys
- [ ] use better generators for more readable data where appropriate
- [ ] new user should be able to run script and see entities loaded in their SOB app

rcullito:
postgresseeduserview
Attached is a screenshot of viewing the UI after running this script a few times. 
Still neeed to: 

- [ ] update the readme
- [ ] fix a bug with USD not saving to postgres

martinklepsch:
Depending on how easy this would be it might be worth considering also adding some SNT/other tokens to some bounties.

src/clj/commiteth/db/seed.clj
```diff
@@ -0,0 +1,193 @@
...
+    ::repository   [{:user_id [::user :id]
+                     :owner [::user :name]}]
+    ::issue        [{:repo_id [::repository :repo_id]}]
+    ::pull_request [{:user_id  [::user :id]
+                     :repo_id  [::repository :repo_id]
+                     :issue_id [::issue :issue_id]}]}))
+
+(def pull-request-result (rs/gen-tree gen-spec relations [::pull_request]))
+
+(def inserted-records (atom []))
```
martinklepsch: As this atom is more part of the script entry point rather than the actual seeding code it might be better to `let` it inside the `seed-postgres` function. Similar points apply to `insert!` I guess.

rcullito: nice points to raise, thanks!

rcullito:
## Testing instructions 

Run `lein run -m commiteth.db.seed/seed-postgres 3` from the root of the project on the command line. 

No other processes need to be running, this task will take care of starting the necessary config and db states via mount. The argument `3` above represents the desired number of entities to create. It will take a minute per entity for each to appear on the page. Once loaded in the app the new bounties should be visible as in the attached screenshot: 
seededtokenbounties

README.md
```diff
@@ -203,6 +203,16 @@ Landing page is static and different CSS and JS due to time constraints.
...
+### Seed postgres for local development
+
+Postgres can be seeded with initial values to make sure that your local SOB app loads with some data. Just
+include the number of entities you wish to create as the first argument to `seed-postgres`, such as:
+
+```
+lein run -m commiteth.db.seed/seed-postgres 3
+```
+
+*Note, it may take a minute for all of your newly created entities to become visible in the local app.
```
martinklepsch: Is this because of reloading? It seems a bit arbitrary without more context. Maybe we can say:

> Note that the app reloads bounties periodically, if you want to see the newly created bounties immediately, just refresh the page :)

martinklepsch: Please add a link to this to the `PostgreSQL` section of the README.

project.clj
```diff
@@ -50,7 +50,10 @@
                  [akiroz.re-frame/storage "0.1.2"]
                  [cljsjs/chartjs "2.6.0-0"]
                  [org.web3j/core "2.3.1"]
-                 [cljs-web3 "0.19.0-0-7"]]
+                 [cljs-web3 "0.19.0-0-7"]
+                 [reifyhealth/specmonstah "1.0.0"]
+                 [clojure-future-spec "1.9.0-beta4"]
```
martinklepsch: We're on 1.9.0 now so we probably no longer need this dependency.

rcullito: nice catch! thanks

README.md
rcullito: yikes, yea that is an arbitrary explanation. I will look into the exact cause and try and be more specific :)

and 👍 on adding to the `PostgreSQL` part of the readme.

rcullito:
This is a tough decision but I feel that since so much of the meaningful state in the SOB Repo is created via Github, the postgres seeding (while nice for initial understanding and visualizing bounties in the app itself) would ultimately be more misleading than helpful for future contributors (i.e. creating a false sense of security regarding data). Going to close this for now unless there are any major objections.

PR #321, opened 2018-02-22
improve exception when contract cannot be loaded

martinklepsch:
This seemingly happens when Geth has not fully synced.

This assert hopefully makes it easier to diagnose issues compared to the
NPE that comes up if the assert is missing.


churik:
Can be built and deployed.
Nothing testable.

PR #323, opened 2018-02-22
swap secretary with bide, move routing into extra ns

martinklepsch:
We've been using secretary but it hasn't been integrated into the codebase very
well. For instance clicking on "Activity" in the menu would not change the url
to include the `#/activity` bit.

This PR fixes that by replacing Secretary with [bide](https://github.com/funcool/bide).
I hope in the future we can extend our usage of URL routing to support page numbers and
other parameters (e.g. sorting) (see #355). This will make it easier for people to share links to
different views of the application.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -5,7 +5,7 @@
                   :exclusions [joda-time]]
                  [re-frame "0.10.2"]
                  [cljs-ajax "0.7.2"]
-                 [secretary "1.2.3"]
+                 [funcool/bide "1.6.0"]
                  [reagent-utils "0.2.1"]
                  [reagent "0.7.0"]
                  [org.clojure/clojurescript "1.9.946"]
--- a/src/cljs/commiteth/core.cljs
+++ b/src/cljs/commiteth/core.cljs
@@ -1,10 +1,10 @@
 (ns commiteth.core
   (:require [reagent.core :as r]
             [re-frame.core :as rf]
-            [secretary.core :as secretary]
             [goog.events :as events]
             [goog.history.EventType :as HistoryEventType]
             [commiteth.ajax :refer [load-interceptors!]]
+            [commiteth.routes]
             [commiteth.handlers]
             [commiteth.subscriptions]
             [commiteth.activity :refer [activity-page]]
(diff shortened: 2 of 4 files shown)


churik:
## 1. Cannot come to 'Manage payouts' screen from 'My payment details'

Steps:
1) open https://testing.openbounty.status.im/app
2) log in
3) open 'Manage Payouts'
4) go to 'My payment details'
5) click 'Manage payouts'

### Actual result:
4) URL changes to `#/my-payment-details`
5) will be redirected to `#/manage-payouts screen`
### Expected result: 
4) URL stays the same (`#/manage-payouts`)
5) user stays on `my-payment-details` screen

**Video:** http://take.ms/cUAtC (first 7 sec - right behavior on `staging`, than actually issue on `testing` env)
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

martinklepsch:
@churik very attentive of you 🙌 Fixed here: https://github.com/status-im/open-bounty/pull/323/commits/445c6442b01c09acaa5658a6e74790f88aa5f428

churik:
@siphiuel can you review it before I proceed with testing?

martinklepsch:
> println to JS console when routes change - do we want to keep it?

No strong opinions, seems fine to keep it around? 

> There is no path when Bounties page is selected. It is assumed to be the default. Should we keep it that way or redirect to /bounties?

For this PR lets stick to whatever was there before. "Redirect" (quotes because it's not HTTP redirect) would be possible though.

> There is an extra hash # in the URL, e.g. /app#/manage-payouts, can we remove it?

Short answer: not for now.    
Long answer: There are two ways to do client side routing URLs, one is using anchors and the other one is using plain URL paths (i.e. no `#`). For plain URL paths the server would need to send back the some app skeleton for each possible URL (e.g. `/app/manage-payouts`) which is certainly possible but doesn't add a lot of value unless you're doing server side rendering. Hope that makes sense.

siphiuel:
I concur. All good then!

churik:
### Environments

### Tested functionality
* Routing through all internal pages;
* external links (in footer)

PR #364, opened 2018-03-20
throw if eth-rpc returns error

martinklepsch:


churik:
PR provides only additional log info, error for which is hard to emulate.
### Tested functionality
* Creating new issue with bounty;
* Successful deployment;

PR #365, opened 2018-03-20
Fix some outdated tests

martinklepsch:
- update an outdated test
- disable a test that requires gas to be in the wallet


churik:
Nothing testable.
Can be built and deployed.

PR #370, opened 2018-03-21
Don't show `Manage Payouts` unless user has issued bounties

martinklepsch:


EugeOrtiz:
IMO it is odd to make options appear/disappear without explaining why this happens. For the new dashboard we can keep the option and on empty state we can encourage to create her/his first bounty.

martinklepsch:
@EugeOrtiz I agree that it should be communicated when/why the menu item appears but at the same time there currently is zero utility of that part of the application for bounty hunters and they will likely never see it. In contrast it might even confuse users into thinking that page has anything to do with getting paid etc.

I'd suggest we just go ahead in the spirit of "moving fast" and reconsider this as soon as we have a better plan. What do you think?

EugeOrtiz:
Yes, its OK now. Just to consider for dashboard V2

churik:
### Environments
* Browser: Chrome, Firefox 58

### Tested functionality
* No `Manage Payouts` if user hasn't any posted bounty
* `Manage payouts`   is displayed when user has at least one posted bounty
* `Manage payouts` appears if user posts 1 bounty

PR #373, opened 2018-03-25
[FIX #355] persist bounty filter to query string

### LINKED ISSUE #355: Persist bounty filters in URL
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story

As a user, I want to copy a link so that visitors are taken directly to the search query I have filtered for on the open bounties page.

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: I would like to be able to create a search filter e.g. owner: aragon on the open bounties page and then share this search filter with someone using a custom link. That way when someone clicks the link they see all of the bounties with that search filter without having to enter the search filter themselves.

#### Expected behavior

- Bounty filters should be represented in the URL as query params
- All current filtering combinations should be correctly encoded in the URI
- Loading a page with filters specified in the URL query params should cause the app to properly load those filters into the application state and update the UI accordingly.

(added by @martinklepsch)

#### Actual behavior

- Filters are not persisted in URL and reloading a page resets all filters.

### Reproduction
[comment]: # (Describe how we can replicate the bug step by step.)

- Enter a search filter e.g. owner: aragon
- Notice that the link remains static and does not update to reflect the search filter

### Solution

- Use [bide query params](https://github.com/funcool/bide#user-guide) to store filters in the application URL.
- Use anchor-style URLs as already done for other routes (i.e. `#/bounties?max-amount=400`).

msuess:
fixes #355 

Persist bounty filters to query string using bide query params as described in #355.

This needs additional testing, as I could not get the github integration to work locally yet.

The filter UI works as expected, but without actual bounties in the DB it's hard to tell if it really works.

@martinklepsch do you think you could take a look at this?

### DIFF
--- a/src/cljs/commiteth/handlers.cljs
+++ b/src/cljs/commiteth/handlers.cljs
@@ -13,7 +13,8 @@
              :as rf-storage
              :refer [reg-co-fx!]]
             [commiteth.ui-model :as ui-model]
-            [commiteth.common :as common]))
+            [commiteth.common :as common]
+            [commiteth.routes :as routes]))
 
 
 (rf-storage/reg-co-fx! :commiteth-sob {:fx :store
@@ -40,6 +41,20 @@
    (println "redirecting to" path)
    (set! (.-pathname js/location) path)))
 
+(reg-fx
+ :persist-bounty-filters-in-query
+ (fn [{:keys [bounty-filters]}]
+   (let [query
+         (->> bounty-filters
+              (remove (comp nil? val))
+              (map (fn [[k v]]
+                     [(ui-model/bounty-filter-type->query-param k)
+                      (ui-model/bounty-filter-value->query-param k v)]))
+              (into {}))]
+     (routes/nav! :bounties {} (if (= {} query)
+                                 nil
+                                 query)))))
+
 (reg-event-fx
  :initialize-db
  [(inject-cofx :store)]
@@ -65,10 +80,17 @@
 
 (reg-event-db
  :set-active-page
- (fn [db [_ page]]
+ (fn [db [_ page params query]]
    (assoc db :page page
              :page-number 1
-             ::db/open-bounties-filters {}
+             ::db/open-bounties-filters
+             (reduce-kv
+              #(let [type (ui-model/query-param->bounty-filter-type %2)]
+                 (assoc %1
+                  type
+                  (ui-model/query-param->bounty-filter-value type %3)))
+              {}
+              query)
              ::db/open-bounties-sorting-type ::ui-model/bounty-sorting-type|most-recent)))
 
 (reg-event-db
@@ -468,9 +490,12 @@
     (merge db {::db/open-bounties-sorting-type sorting-type
                :page-number 1})))
 
-(reg-event-db
-  ::set-open-bounty-filter-type
-  (fn [db [_ filter-type filter-value]]
-    (-> db
-        (assoc-in [::db/open-bounties-filters filter-type] filter-value)
-        (assoc :page-number 1))))
+(reg-event-fx
+ ::set-open-bounty-filter-type
+ (fn [{:keys [event db]} [_ filter-type filter-value]]
+   (println "db" db)
+   (let [filters (::db/open-bounties-filters db)]
+     (println "filters" filters)
+     {:persist-bounty-filters-in-query
+      {:bounty-filters
+       (assoc filters filter-type filter-value)}})))
--- a/src/cljs/commiteth/routes.cljs
+++ b/src/cljs/commiteth/routes.cljs
@@ -14,12 +14,17 @@
   "A function which will be called on each route change."
   [name params query]
   (println "Route change to: " name params query)
-  (rf/dispatch [:set-active-page name]))
+  (rf/dispatch [:set-active-page name params query]))
 
 (defn setup-nav! []
   (bide/start! router {:default     :bounties
                        :on-navigate on-navigate}))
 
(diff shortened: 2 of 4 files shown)


src/cljs/commiteth/handlers.cljs
```diff
@@ -39,6 +40,20 @@
...
+   (let [query
+         (reduce-kv
+          #(if (nil? %3)
+             %1
+             (let [param-name (ui-model/bounty-filter-type->query-param %2)
+                   param-value (ui-model/bounty-filter-value->query-param %2 %3)]
+               (assoc %1 param-name param-value)))
+          {}
+          bounty-filters)]
+     (routes/nav! :bounties {} query))))
```
martinklepsch:
```clojure
;; untested
(->> bounty-filters
     (remove (comp nil? val))
     (map (fn [[k v]]
            [(ui-model/bounty-filter-type->query-param k)
             (ui-model/bounty-filter-value->query-param k v)]))
     (into {}))
```
Maybe something like this is a bit easier to read? Also no need to bind via `let` if the value is only used once.

src/cljs/commiteth/routes.cljs
```diff
@@ -14,12 +14,18 @@
...
   (bide/start! router {:default     :bounties
                        :on-navigate on-navigate}))
 
-(defn nav! [route-id]
-  (bide/navigate! router route-id {}))
+(defn nav!
+  ([route-id]
+   (nav! route-id {}))
+  ([route-id params]
+   (nav! route-id params {}))
```
martinklepsch: Use `nil` instead of empty maps here to avoid stale `?` at the end of the URI.

msuess:
Thanks for your feedback @martinklepsch, I'll be looking into this tonight!

src/cljs/commiteth/handlers.cljs
martinklepsch: Actually regarding use of `let` — sometimes it's good to explicitly declare what something is. So if you feel like binding stuff to a name here adds value then please keep it :)

msuess: Keeping the `let` but using your implementation. I was pretty unhappy with what I came up with. This is so cool, thanks for the feedback!

src/cljs/commiteth/routes.cljs
msuess: Done!

msuess:
@martinklepsch I fixed the stuff you commented on. Regarding calling `nav!` in the UI components: I went down this path for a little while but hit a road block: in order to *add* to the query string, one needs to have access to all the current filters when calling `nav!`. This is actually one of the reasons that made me move this logic into an fx handler, now that I think of it. Do you have any ideas as to how to solve this?

msuess:
@martinklepsch there are no changes in the UI code, since the call to `nav!` is still in the handlers.

martinklepsch:
@msuess I see what you mean. Did not anticipate that properly, sorry for leading you into that 🙂 Dabbling with this for some time lead me to some other ideas to improve this code but we probably shouldn't be tackling these as part of this PR/bounty. Some notes below just to keep record.

---
Basically we're duplicating the information in the query params in `(::db/open-bounties-filters db)`. Duplication like this is problematic because different parts of the app might rely on this duplication, causing pain when refactoring. 

##### Idea for a solution

Ultimately that key `::db/open-bounties-filters` in the DB should be replaced by a subscription which derives the same information from the query params. These are currently not stored in the DB but you could imagine that the `[page params query]` tuple is stored in the DB (e.g. under a `:route` key). This would be trivial to add to `:set-active-page`.

The subscription would then also be responsible for the transformation of that data into the format the UI components expect (i.e. `ui-model/query-param->bounty-filter-type`). 

`::set-open-bounty-filter-type` (which gets `db` as arg) would simply extract the current routing information from the DB (query params) and call `nav!` with the updated query params (basic `assoc`).

As you rightfully noted it's problematic to call `nav!` in the components themselves because that requires them to have a full understanding of the current routing state.

Don't worry if this doesn't make sense right now, jotting this down in a bit of a rush. 😅

msuess:
@martinklepsch fixed the issue with the currencies filter. the problem was that the value for ETH was a string and the other token-ids are keywords. I changed the value for ETH to a keyword as well, let me know if you prefer strings for everything :)

martinklepsch:
Just played with this again and noticed that the owner selection now has a problem: 

- when clicking on an owner 
  - the query params are changed
  - the bounties are filtered properly
  - **but** the checkbox does not become enabled
  - (perhaps a consequence) when clicking the same names again sometimes the same owner is listed twice in the query params

msuess:
@martinklepsch sorry, I should have catched that. Will fix.

martinklepsch:
No worries! 👍

churik:
I addition to @martinklepsch comment: when applying the filter by the `Owner`, the bounties are not filtered (always 'No matching bounties found.')

**Video:** http://take.ms/PwIxp

msuess:
@martinklepsch @churik fixed the issue when filtering by owner.

churik:
=====TEMPLATE

### Environments
* Operating System: macOS High Sierra 10.13.3 (Chrome, Firefox 58, Safari 11), Windows 10 (EDGE 10, IE 11, Chrome 64)

### Tested functionality
* All filters can be applied, all are persisted
* Copy-paste url in different browsers - filters applied

PR #394, opened 2018-04-06
Add less to Dockerfile

siphiuel:
Add `less` to Dockerfile. Simplifies debugging when e.g. it is necessary to view app logs inside container.


martinklepsch:
Sure, let’s do it then :+1:

PR #395, opened 2018-04-06
fix DR-0001 link

rcullito:
I came across this while browsing around prior decisions. Love the concept, just updating a link in the readme 👍.


martinklepsch:
Thanks, good catch!

PR #400, opened 2018-04-11
Some watch-token improvements

siphiuel:
- Re-execute watch if max attempt count exceeded;
- block multiple execute tx attempts in case when there are funds still on account (probably due to missing watch invocation);
- invoke watch synchronously prior to confirming transactions


siphiuel:
Obsoleted by #447

PR #402, opened 2018-04-12
WIP: allow posting of issue comments from github account with2FA enabled

rcullito:
My local bot account was blocked from commenting on a labelled issue with a bounty by github's api. Per Github's documentation: https://developer.github.com/v3/auth/#working-with-two-factor-authentication, this pull request aims to set the X-GitHub-OTP header. I am pretty sure it should come from a personal access token created through github (instructions here: https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) rather than being one of the access tokens we are receiving back within the app itself. Although suggestions are welcome around process here :)

<blockquote><div><strong><a href="https://developer.github.com/v3/auth/#working-with-two-factor-authentication">Other Authentication Methods | GitHub Developer Guide</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/">Creating a personal access token for the command line - User Documentation</a></strong></div><div>You can create a personal access token and use it in place of a password when performing Git operations over HTTPS with Git on the command line or the API.
…</div></blockquote>

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ github-client-id | Related to OAuth. Copied from GitHub account Settings->Develo
 github-client-secret | Related to OAuth. Copied from GitHub account Settings->Developer settings->OAuth Apps
 github-user | GitHub username for bot account. It is used for posting bounty comments
 github-password | GitHub password for bot account
+github-token | GitHub token for bot account. (see
+https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/ for details)
 webhook-secret | Secret string to be used when creating a GitHub App
 user-whitelist | Set of GitHub user/org IDs to be whitelisted. E.g. `#{"status-im" "your_org"}`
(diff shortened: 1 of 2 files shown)


martinklepsch:
Why did you enable 2FA on a bot account? Not sure but maybe it's fine if the bot account has to have 2FA disabled for now? What do you think

rcullito:
After talking with other members of the team it seems like it is common practice to have a separate account for the bot. Taking a closer look at github's v3 api response, (specifically the X-GitHub-OTP response header which was still mentioning SMS) made me realize this is not an ideal use case to plan for, or at least not one with much upside. Rather than close this, I'm going to look at just updating the documentation to point to the value in creating a separate bot account w/o 2FA enabled to so that future contributors don't get tripped up over it.

PR #406, opened 2018-04-12
[Fix #259] Migrate to DB backed sessions

### LINKED ISSUE #259: Save web sessions state in a service
### User Story

As a devops I want to restart the web server or change it to another server without it losing web sessions state.

### Description

Right now it's being used `immutant.web.middleware/wrap-session` to save the state.

The expected behaviour is to save the state in Postgres, so that we can just not add more kinds of services for now. It could be done with https://github.com/luminus-framework/jdbc-ring-session

The table should be created when starting the app if it doesn't exist.

Some docs about this topic: http://www.luminusweb.net/docs/sessions_cookies.md

martinklepsch:
> This is a copy of #288 because we couldn't update the contributors branch.

Fixes https://github.com/status-im/open-bounty/issues/259
Status: Finished

Summary
This migrates from Immutant servlet based sessions to DB backed sessions using
[ring.middlware.session](https://ring-clojure.github.io/ring/ring.middleware.session.html),
[jdbc-ring-session](https://github.com/luminus-framework/jdbc-ring-session) and
[ring-session-timeout](http://ring-clojure.github.io/ring-session-timeout/)

Why?
- Need to potentially store more data in the session.
- Need to share session information across different application processes.
- A user's session should survive application restarts.

Things to note:
- A new table `session_store` has been added to hold session information. This
will be automatically created when the app starts.
- Any stale session records will be automatically deleted by a session cleaner
that runs on the background. A session is considered stale after 6 hours of
inactivity.

Alternatives:
I think this might become a bottleneck in the future as every new session hits
the DB, but not really sure about this as I don't know the current usage patterns.
An alternative could be to use encrypted cookie based sessions, but this depends
on the data that's goint to be stored in the session.

I'd appreciate your feedback on the approach as well as on the implementation.

More info:

- http://www.luminusweb.net/docs/sessions_cookies.html
- https://github.com/ring-clojure/ring/wiki/Sessions

<blockquote><div><strong><a href="http://www.luminusweb.net/docs/sessions_cookies.html">Luminus - a Clojure web framework</a></strong></div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/ring-clojure/ring">ring-clojure/ring</a></strong></div><div>ring - Clojure HTTP server abstraction</div></blockquote>

### DIFF
--- a/env/dev/resources/config.edn
+++ b/env/dev/resources/config.edn
@@ -50,5 +50,6 @@
 
  :user-whitelist #{}
 
+ :session-timeout 21600
  ;; used for blacklisting tokens from token registry data
  :token-blacklist #{}}
--- a/env/prod/resources/config.edn
+++ b/env/prod/resources/config.edn
@@ -4,4 +4,5 @@
  :nrepl-port        7000
  :server-address    "https://commiteth.com"
  :html2png-command  "/home/commiteth/html2png.sh"
+ :session-timeout   21600
 }
--- a/env/test/resources/config.edn
+++ b/env/test/resources/config.edn
@@ -1,4 +1,5 @@
 {
  :test              true
  :jdbc-database-url "jdbc:postgresql://localhost/commiteth?user=commiteth&password=commiteth"
+ :session-timeout   21600
 }
--- a/project.clj
+++ b/project.clj
@@ -18,6 +18,8 @@
                  [ring/ring-json "0.4.0"]
                  [ring/ring-defaults "0.3.1"]
                  [ring/ring-codec "1.0.1"]
+                 [ring/ring-session-timeout "0.2.0"]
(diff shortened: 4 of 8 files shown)


churik:
It is not working for me at all.

Steps: 
- deploy PR-406 to `testing env`
- edit `config.edn` in `testing.openbounty.status.im` vm (set `:session-timeout`  to  180}
- restart docker container
- log in to https://testing.openbounty.status.im, wait for 200 sec. (I worked with other tabs)
- refresh  https://testing.openbounty.status.im

## Expected Result
I was logged out
## Actual Result
I wasn't logged out.

Or I understand inactivity in wrong way, or missed smth in configuration, but any case I cannot make it works.

PR #407, opened 2018-04-12
improved logging in commiteth.scheduler

martinklepsch:


src/clj/commiteth/scheduler.clj
```diff
@@ -361,16 +357,15 @@
                                    token-balances)
             (update-issue-usd-value contract-address))))
       (catch Throwable ex 
-        (do (log/error "update-balances exception:" ex)
-            (clojure.stacktrace/print-stack-trace ex))))))
```
martinklepsch: @siphiuel @rcullito FYI `log/error` supports receiving a `Throwable` as first arg and will properly print it's stacktrace etc.

PR #415, opened 2018-04-13
Throw exception if about to use an old nonce

siphiuel:
`get-signed-tx` is used by:
1. `multisig/create-new`
2. `multisig/watch-token`
3. `multisig/send-all`

If an exception is thrown, then
1. `create-new` (contract deployment) will be retried by `deploy-pending-contracts` scheduler thread
2. `watch-token` will be retried by `update-bounty-token-balances` scheduler thread
3. `send-all` will be retried by `self-sign-bounty` scheduler thread

### DIFF
--- a/src/clj/commiteth/eth/core.clj
+++ b/src/clj/commiteth/eth/core.clj
@@ -65,7 +65,6 @@
         (if (= nonce @current-nonce)
           (throw (Exception. (str "Attempting to create transaction with the same nonce: " nonce)))
           (swap! current-nonce (constantly nonce)))
-        (log/info "Current nonce:" nonce)
         nonce))))
 
 (defn get-signed-tx [gas-price gas-limit to data]
--- a/src/clj/commiteth/eth/core.clj
+++ b/src/clj/commiteth/eth/core.clj
@@ -53,18 +53,27 @@
           (throw (ex-info "Make sure you provided proper credentials in appropriate resources/config.edn"
                           {:password password :file-path file-path}))))))
 
+(defn get-nonce []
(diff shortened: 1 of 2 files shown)


src/clj/commiteth/eth/core.clj
```diff
@@ -53,18 +53,26 @@
...
    encoded in credentials.
    See https://web3j.readthedocs.io/en/latest/transactions.html#offline-transaction-signing"
-  (let [web3j (create-web3j)
-        nonce (.. (.ethGetTransactionCount web3j 
-                                           (env :eth-account) 
-                                           DefaultBlockParameterName/LATEST)
-                  sendAsync
-                  get
-                  getTransactionCount)
+  (let [nonce ((get-nonce))
```
martinklepsch: Given that you re-create the inner `get-nonce` function every time `get-signed-tx` is called the exception condition `(= nonce @current-nonce)` will never be reached I think?

martinklepsch:
I think an alternative (and harder to implement) approach here might be maintaining some state of an internal transaction identifier (e.g. `bounty-contract-$issue-id`) to `nonce`. This way we could send of multiple transactions concurrently without waiting for the nonce to be increased by web3j.

src/clj/commiteth/eth/core.clj
siphiuel: Right, what a lame mistake( Fixed

churik:
### Environments

### Tested functionality
* 20 different contracts are successfully deployed (added bounty label to issues at the same time) ;
* readd bounty label - no new contract is deployed;
* claiming, merging, confirming payout

PR #427, opened 2018-04-17
Send txs sequentially; add thread to check for unmined txs

siphiuel:
This PR does the following:
  - do not send txs until previous one has been mined or discarded
  - check for stale txs in check-tx-receipts thread
 - monitor tx execution via `eth/track-tx!` and `eth/untrack-tx!`

Some additional refactors have been made:
  - add a multimethod for `eth/execute` dispatched on the value of `offline-signing?`
  - replace `update-[name]-hash` with a generic update fn
  - return `tx-info` maps from `eth/execute` fn. These are afterwards passed to `eth/[un]track-tx!` fns for DB/in-memory storage

### After QA:
Small issues to fix:
- [x]  too small timeout which triggers transaction to re-execute 
- [x]  two comments are persisted if 1 transaction was canceled because of timeout
- [ ]  'Deploying contract' - usually takes 1-2 min - is too optimistic


src/clj/commiteth/eth/core.clj
```diff
@@ -51,59 +52,38 @@
...
+  (let [nonce (.. (.ethGetTransactionCount @web3j-obj 
+                                           (env :eth-account) 
+                                           DefaultBlockParameterName/PENDING)
+                  sendAsync
+                  get
+                  getTransactionCount)]
+    (if (or (nil? @nonce-being-mined)
+            (> nonce (:nonce @nonce-being-mined)))
+      (:nonce (reset! nonce-being-mined (TxNonce. nil nonce nil nil)))
+      (throw (Exception. (str "Attempting to re-use old nonce" nonce))))))
```
martinklepsch: This function is doing more than getting a nonce now, consider renaming it or moving the mutating code somewhere else.

src/clj/commiteth/eth/multisig_wallet.clj
```diff
@@ -43,6 +43,7 @@
   [{:keys [owner internal-tx-id]}]
   {:pre [(string? owner) (string? internal-tx-id)]}
   (eth/execute {:internal-tx-id internal-tx-id
+                :type "deploy"
```
martinklepsch: Maybe instead of introducing another parameter here we could modify `internal-tx-id` to look something like this:

```clojure
[:deploy "1231455"]
```

resources/sql/queries.sql
```diff
@@ -145,7 +145,8 @@ RETURNING repo_id, issue_id, issue_number, title, commit_sha, contract_address;
 -- :name update-transaction-hash :! :n
 -- :doc updates transaction-hash for a given issue
 UPDATE issues
-SET transaction_hash = :transaction_hash
+SET transaction_hash = :transaction_hash,
+    updated = timezone('utc'::text, now())
```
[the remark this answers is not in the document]
siphiuel: Agree with introducing separate columns for timestamps, but perhaps not in this PR. It would be perhaps nice to have a separate table, something like `tx_data(hash, type, updated, issue_id)` and link it to `issues`. What do you think?

src/clj/commiteth/eth/core.clj
```diff
@@ -51,59 +52,38 @@
...
+             nonce
+             gas-price
+             gas-limit
+             to
+             data)
+        signed (TransactionEncoder/signMessage tx (creds))
+        hex-string (Numeric/toHexString signed)]
+    (log/infof "Signing TX: nonce: %s, gas-price: %s, gas-limit: %s, data: %s"
+               nonce gas-price gas-limit data)
+    hex-string))
```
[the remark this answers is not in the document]
siphiuel: Agree, fixed.

src/clj/commiteth/eth/core.clj
```diff
@@ -173,7 +153,7 @@
     (cond
       ;; Ignore any responses that have mismatching request ID
       (not= (:id result) request-id)
-      (log/error "Geth returned an invalid json-rpc request ID, ignoring response")
+      (throw (ex-info "Geth returned an invalid json-rpc request ID, ignoring response"))
```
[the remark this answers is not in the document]
siphiuel: Exception are always caught up the call stack, for `deploy`, `execute`, and `watch`.

src/clj/commiteth/eth/core.clj
```diff
@@ -263,8 +243,13 @@
...
   [hash]
-  (eth-rpc {:method "eth_getTransactionReceipt"
-            :params [hash]}))
+  (when-let [receipt (eth-rpc {:method "eth_getTransactionReceipt"
+                               :params [hash]})]
+    (if (= hash (:tx-hash @nonce-being-mined))
+      (reset! nonce-being-mined nil)
+      ;; This can happen in case of payout receipts
+      (log/infof "Obtained receipt for tx not currently being mined: " hash))
+    receipt))
```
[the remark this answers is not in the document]
siphiuel: Agree, relieved it of extra functionality.

src/clj/commiteth/eth/core.clj
siphiuel: Renamed.

src/clj/commiteth/eth/multisig_wallet.clj
siphiuel: Great idea, have done so.

resources/sql/queries.sql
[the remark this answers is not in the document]
martinklepsch: Let's capture that in an issue so we don't forget

src/clj/commiteth/eth/core.clj
```diff
@@ -51,59 +53,94 @@
...
-  "Create a sign a raw transaction. 'From' argument is not needed as it's already encoded in credentials.
-   See https://web3j.readthedocs.io/en/latest/transactions.html#offline-transaction-signing"
-  (let [nonce (get-nonce nonce-tracker internal-tx-id)]
-    (log/infof "%s: Signing nonce: %s, gas-price: %s, gas-limit: %s"
-               internal-tx-id nonce gas-price gas-limit)
-    (-> (RawTransaction/createTransaction (biginteger nonce) gas-price gas-limit to data)
-        (TransactionEncoder/signMessage (creds))
-        (Numeric/toHexString))))
+
+(defprotocol TxTracker
```
martinklepsch: I like to prefix protocols with `I` so it's more apparent when they appear in stack traces. Just for your consideration :)

src/clj/commiteth/eth/core.clj
```diff
@@ -51,59 +53,94 @@
...
+
+(defn track-tx! 
+  "Store tx data in tx-tracker and DB"
+  [{:keys [issue-id tx-hash type]
+                  :as tx-info}]
+  (track-tx tx-tracker tx-info)
+  (issues/save-tx-info! issue-id tx-hash type))
+  
+(defn untrack-tx! 
+  "Mark tx data stored in tx-tracker and DB as closed"
```
martinklepsch: `closed` is used in a few places here. Maybe `successfully mined` or `confirmed` would be more specific here?

src/clj/commiteth/eth/core.clj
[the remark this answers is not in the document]
martinklepsch: let's add the `result` as exception data here. (`ex-info` expects two arguments in general)

src/clj/commiteth/eth/core.clj
```diff
@@ -277,36 +316,68 @@
...
+        value (format "0x%x" 0)]
+    (cond-> {:data data
+             :from  from
+             :value value}
+      gas-price
+      (merge {:gasPrice (integer->hex gas-price)})
+      contract
+      (merge {:to contract}))))
+
+(defmulti execute (fn [arg] (offline-signing?)))
```
martinklepsch: Using a multimethod with boolean is a bit unusual, for readability it might be better to make the dispatch fn return some more descriptive value:

```clojure
(if (offline-signing?) :offline :whatever-youd-call-the-other-thing)
```

Alternatively you could just have something like

```clojure
(defn execute-offline ,,,)

(defn execute-bla ,,,)

(defn execute ,,,
  (case (offline-signing?)
    true (execute-offline ,,,)
    false (execute-blabla ,,,)))
```

Multimethods shine when you're trying to create an open extension mechanism, that is you want to allow other people to provide additional implementations of this functionality. In this case I'm not sure how likely it is that we will do this. 

Definitely nothing wrong with using a multimethod here though! :+1:

src/clj/commiteth/scheduler.clj
```diff
@@ -52,11 +52,14 @@
          (when-let [receipt (eth/get-transaction-receipt transaction-hash)]
            (log/infof "issue %s: update-issue-contract-address: tx receipt: %s" issue-id receipt)
            (if-let [contract-address (multisig/find-created-multisig-address receipt)]
-             (let [issue   (issues/update-contract-address issue-id contract-address)
+             (let [_ (eth/untrack-tx! {:issue-id issue-id 
+                                           :tx-hash transaction-hash 
+                                           :result contract-address 
+                                           :type :deploy})
```
martinklepsch: indentation

resources/sql/queries.sql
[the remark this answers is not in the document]
siphiuel: Created a new issue for this: https://github.com/status-im/open-bounty/issues/442

src/clj/commiteth/eth/core.clj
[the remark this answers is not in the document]
siphiuel: Added.

src/clj/commiteth/eth/core.clj
siphiuel: Renamed.

src/clj/commiteth/eth/core.clj
siphiuel: Substituted with `mined`.

siphiuel:
QA notes:
1.update issues set transaction_hash='0x6c1fe9f2f3a1d948ec49fd721aae60b2b5d3378b007d668d55cd2d334f13cf04', contract_address=null, updated=updated - interval '15 minutes' where issue_id=318430379;

churik:
## 1. Too small timeout which triggers transaction to re-execute (`Mainnet`)

Steps:
- create several issues (8 in my case)
- put bounty label an all issues
### Expected result:
all transactions are mined successfully
### Actual result:
3 of 8 transactions are canceled because of timeout (currently 5 min)

## 2. Two comments are persisted if 1 transaction was canceled because of timeout 

Steps:
- open any issue with canceled transaction (i.e. https://github.com/TestingEnvENV/TestEnvAtt5_1/issues/227)
### Expected result:
1 comment (unsuccessful comment from bot will be replaced with successfull) 
### Actual result:
2 comments are persisted

## 3. 'Deploying contract' - usually takes 1-2 min - is too optimistic
I suggest to replace it with 3- to timeout time(IMO 10 min)
d

churik:
### Tested functionality

* Creating several issues with `bounty` label (6-10 items simultaniously; contract is deployed one-by-one);
* Full workflow for 1 bounty (created -> funded -> claimed -> PR merged -> funds transfer to BHunter)
* Re-deploy contract

Tested on Ropsten and Mainnet.

Comment **3.  'Deploying contract' - usually takes 1-2 min - is too optimistic**  depends on network speed and should be checked on prod and discussed separately, probably will move to the separate issue.

churik:
@martinklepsch please merge if it is OK after review; QA is passed.

src/clj/commiteth/eth/token_registry.clj
```diff
@@ -1,6 +1,7 @@
 (ns commiteth.eth.token-registry
   (:require [commiteth.eth.core :as eth]
             [commiteth.config :refer [env]]
+            [commiteth.eth.web3j :refer [web3j-obj creds-obj]]
```
martinklepsch: Not important but I generally avoid `:refer` as it makes it harder to see where stuff is coming from.

Issue #284, opened 2018-02-14
Bounty revocation: UI + backend


siphiuel:
*Type*: Feature

### Description
Currently, bounties cannot be revoked and funds cannot be withdrawn from bounties assigned to issues that are no longer considered to be up for a bounty.

Therefore, it would be nice to be able to remove GitHub bot comment and withdraw funds from associated contract if an issue had a `bounty` label removed. Thus revoked bounties should be somehow visible in SOB UI.

### Solution
On the UI side, add a section named "Revoked Bounties" to the "Manage Payouts" screen. On the backend, extend #235 with the revocation logic, as currently #235 will only remove bounties when there are no contract funds. This means that once a `bounty` label has been removed, contract funds will be withdrawn to the repo owner's address.

pablodip:
The title of the issue only mentions UI, and the description also backend and new logic. How about splitting those in different issues?

Regarding the UI, it should probably be defined and design with @arash009 and @denis-sharypin, shouldn't it?

pablodip:
I didn't mean to fix the title. :)) But to split the issues so that we can treat them better separately, since even though they are about revocation, they are different things.

For example we need to talk to @3esmit and @andytudhope about what we talked in point 3.) in issue #234.

And with @arash009 and @denis-sharypin the UI thing.

siphiuel:
I think it's fine to keep it as a single issue. It can implement some basic UI, and then as suggested earlier we can improve UI separately. Having some kind of UI in the basic package will also simplify testing I guess. Just my thoughts.

pablodip:
1.) How are revoked issues shown right now in the UI? Or not shown at all?
2.) How are revokes issues marked in the database/domain right now?

siphiuel:
1) Not shown at all
2) If/when #235 is merged, then revoked issues will reside in the `archive` table

martinklepsch:
Important part of this would also be cleaning up production data. :+1:

rcullito:
relevant context here: https://github.com/status-im/open-bounty/issues/234

PR #430, opened 2018-04-19
bounty revocation

rcullito:
Part of https://github.com/status-im/open-bounty/issues/284

The workflow right now is to hit a `revoke` button on the manage payouts page which triggers a series of steps which are better chronicled here: https://realtimeboard.com/app/board/o9J_kz0D-rM=/


martinklepsch:
A few thoughts:
- Do we need to file a fake claim? Could we not simply set `winner_login`? Would we need to set `winner_login` at all? Maybe an alternative would be to just create the payout transaction with the owner's address?
- It's important that we track if a bounty has been revoked, consider adding a column `revoked_ts` with a timestamp of revocation. I see you did that to some extent in the `commit_sha` column but this might warrant an actual column. 

An interesting implication around this is also that multiple people may pay into a bounty. Ideally each of them would get their funds back but I'm not sure how feasible this is at this point.

rcullito:
@martinklepsch thanks for reviewing this!

> Do we need to file a fake claim? Could we not simply set winner_login? 
 If we want this bounty to show up in the "manage payouts" dashboard with the existing logic, then we do need a claim in place. (Or conversely, we could update the logic for how an entity shows up in the manage payouts dashboard 😄) I totally agree that there is a lot of "fat" to be trimmed here for the revocation scenario, but I wanted to witness a few full end to end payouts before starting to cut out the  unnecessary parts. 

> It's important that we track if a bounty has been revoked, consider adding a column revoked_ts with a timestamp of revocation.
👍 yepp, makes sense! 

> multiple people may pay into a bounty
good point! this issue, and a few others like it are causing me to try and research events emitted by the smart contracts themselves, so our source of truth is not derived solely from postgres, which is just the aggregate balances without a breakdown by contributor. I've been checking out this library (https://github.com/district0x/re-frame-web3-fx#web3watch-events) as it seems like it is easier to respond to events via the browser than constantly polling over json-rpc on the backend, but it's totally possible that I'm missing some part of the current sob codebase and we may already be doing more in this area than I understand cc @siphiuel

arash009:
cc @3esmit re the multiple people paying into a contract to provide clarity how this will work with the revised smart contract.

rcullito:
@martinklepsch thanks for mulling it over and agreed that an api w/ issue-id makes the most sense from a UX perspective (as well as the clear intention to revoke). 

The main bottleneck for me is ensuring that the confirm-payout then works for an address(addresses) other than the repo owner as trials so far in that area have been failing at the contract level, although I'm trying to work through it.

contracts/README.md
```diff
@@ -2,8 +2,7 @@
 
 This directory contains all the underlying smart contracts used by the OpenBounty platform.
 
-- A script `contracts/build.sh` is part of this repository and can be used to
-compile the contracts and copy Java interfaces into `src/java/`.
+- Run `lein build-contracts` to compile the contracts and copy Java interfaces into `src/java/`.
```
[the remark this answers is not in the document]
martinklepsch: Good point. There's probably some magic bash trick that would fix this 😀

rcullito:
cc @siphiuel @martinklepsch I still have a couple of UX features to continue improving after chatting with Euge, but was hoping to get your feedback on how the overall implementation looks before fiddling with any further changes :)

What we decided was easiest was the scenario of the user hitting a single button "revoke" to kick off the various steps. Work is then dispatched from a reframe handler to a route that polls for confirmation from the commiteth bot, and on success dispatches a confirm payout event that prompts for confirmation from metamask.

src/clj/commiteth/scheduler.clj
```diff
@@ -255,6 +282,43 @@
     (neg? n) (- n)
     :else n))
 
+(defn contract-confirmation-logs [contract-address]
+  "retrives all log events for the confirmation topic since contract creation"
+  (some-> contract-address
+      issues/get-issue-by-contract-address
+      :transaction_hash
```
siphiuel: Why are we fetching `transaction_hash` instead of `execute_hash`? (not saying this is wrong, just curious about contract log fetching in general)

resources/migrations/20180426153251-update-issue-execution-posted.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "public"."issues" ADD COLUMN "execute_posted" boolean;
```
siphiuel: Where is this field used?

src/clj/commiteth/scheduler.clj
```diff
@@ -255,6 +282,43 @@
...
+      eth/get-transaction-by-hash
+      :blockNumber
```
siphiuel: Also - are we sure that the block containing tx with `:transaction_hash` will also be the one containing tx with `:execute_hash`?

src/clj/commiteth/routes/services.clj
```diff
@@ -237,4 +257,19 @@
...
-                         (ok (user-bounties user))))))
+                         (ok (user-bounties user)))
+                    (POST "/revoke"  {{issue-id         :issue-id
+                                       contract-address :contract-address
+                                       owner-address    :owner-address} :params}
+                          :auth-rules authenticated?
+                          :current-user user
+                          (do (log/infof "calling revoke-initiate for %s with %s %s" issue-id contract-address owner-address)
+                              (if-let [{:keys [execute-hash execute-write]} (execute-revocation issue-id contract-address owner-address)]
+                                (if (scheduler/poll-transaction-logs execute-hash contract-address)
```
siphiuel: Instead of having to synchronously poll here in request handler, can we instead rely on `load-data` fn from `core.cljs` to periodically fetch whether there is a confirm hash for the revocation tx? Fetching confirm hash involves invoking `multisig/find-confirmation-tx-id`, which is very similar thing to what `scheduler/contract-confirmation-logs` does.

Perhaps we could re-use `:load-owner-bounties` event for this purpose (given that `winner-login` field will be set to owner's login so that `owner-bounties` query works as expected)

src/clj/commiteth/scheduler.clj
rcullito: My feeling was that (at least within the context of the revocation workflow) that the execution_hash would still be in a pending block. 

So transaction_hash is used in this case to define a starting point that we can safely assume has already been mined. Then with the derived block number from that transaction there is enough info to poll the minimal amount of event logs to check for some later event. (in the case of revocation what is actually being polled for is the presence of the execution_hash). 

Thanks for highlighting this as I believe it could be made more extensible by allowing the type of hash (transaction, execution, etc...) to be passed in as a param such as `since_hash` or `starting_point` rather than being hardcoded. Will try and keep that in mind if I spot another use case for this type of polling.

resources/migrations/20180426153251-update-issue-execution-posted.up.sql
rcullito: Nice catch, thank you! My original implementation for revocation was perhaps overly explicit in the steps underneath the hood that were being performed. It has since been agreed upon that the UI should contain as few steps as possible, so I should rip out these legacy columns.

src/clj/commiteth/routes/services.clj
rcullito: @siphiuel yes this really is the crux of the PR in my opinion, really good to discuss. 

It seems like the main thing is knowing *when* to call `update-confirm-hash` and there is some similarity between  `scheduler/contract-confirmation-logs` and `multisig/find-confirmation-tx-id` because the former is basically ensuring that the conditions will be met for the latter to return something when it is called in this context 😄 

I agree it would be nice to trim out this excess polling, but if I do I'm still not sure what the most succinct thing is to pay attention to for ensuring that the execute_hash is mined? Was thinking the `tx-tracker` you introduced might come in handy here? 

This is kind of a tricky thing to reason about so let me know if I'm missing something obvious, but the reason I haven't relied on `load-data` or `owner-bounties` yet is that if `update-confirm-hash` is not called manually as listed above, it will not be picked up by the scheduler thread since a revoked bounty has no associated PR (one of the conditions of `db/pending-payouts`). 

Thanks again for checking this out, let me know what you think!

src/clj/commiteth/routes/services.clj
```diff
@@ -155,13 +162,26 @@
...
-    {:swagger {:ui   "/swagger-ui"
-               :spec "/swagger.json"
-               :data {:info {:version     "0.1"
-                             :title       "commitETH API"
-                             :description "commitETH API"}}}})
+    {:swagger    {:ui   "/swagger-ui"
+                  :spec "/swagger.json"
+                  :data {:info {:version     "0.1"
+                                :title       "commitETH API"
+                                :description "commitETH API"}}}
```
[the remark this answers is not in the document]
rcullito: Yepp, definitely a fair point! Thanks for spotting. In general totally agree it obscures the nature of the change: on these kind of abnormally large sexps such as `defapi` even moreso. 

Once it's in though, do you have any advice on unwinding so it doesn't confuse the reviewer, or is the best antidote really just preventing it from happening in the first place?

src/clj/commiteth/routes/services.clj
```diff
@@ -237,4 +257,17 @@
                          :auth-rules authenticated?
                          :current-user user
                          (log/debug "/user/bounties")
-                         (ok (user-bounties user))))))
+                         (ok (user-bounties user)))
+                    (POST "/revoke"  {{issue-id         :issue-id
+                                       contract-address :contract-address
+                                       owner-address    :owner-address} :params}
```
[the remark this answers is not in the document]
rcullito: good point! yes that is correct, issue-id is all that is needed to go fetch the rest 👍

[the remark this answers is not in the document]
rcullito: will trim this down.

src/clj/commiteth/routes/services.clj
rcullito: @siphiuel yes that's a good point. I believe if I created a separate scheduler job to check for receipts of revoked bounties (they won't show up in the current query because there is no associated PR), then it should be possible to just follow the consistent workflow we already have rather than polling from within the request handler. 

I guess my main question then becomes the mechanism by which we would know to either call `load-data` or `owner-bounties` to get the confirm-hash in hand in order to call confirm-payout. Do we have a way to call these handlers periodically to the extent they can't be event driven? (happy to write such a feature – recurring dispatch or something – too, just curious if it exists already for the use case you describe?)

Thanks!

src/clj/commiteth/scheduler.clj
```diff
@@ -248,6 +276,43 @@
...
+    ;; polling will be slow but if we want to move to an event driven
+    ;; model then we can listen for events, rather than logs, once we're
+    ;; using a geth node again
+    (chime-at intervals
+              (fn [time]
+                (when (execution-status execute-hash contract-address)
+                  (deliver found? true)))
+              {:on-finished (fn []
+                              (deliver found? false))})
+    @found?))
```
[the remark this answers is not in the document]
rcullito: Good point! As with a few of the other changes that were introduced this will be addressed by removing this functionality entirely. The big shift that actually vitaliy pointed out is that reframe is already reloading data via a series of api calls every minute, so revocation can take advantage of that rather than having to do its own polling.

src/clj/commiteth/scheduler.clj
```diff
@@ -190,24 +195,8 @@
...
-           owner :owner
-           comment-id :comment_id
-           issue-number :issue_number
-           balance-eth :balance_eth
-           tokens :tokens
-           confirm-id :confirm_hash
-           payee-login :payee_login
-           updated :updated} (db-bounties/confirmed-payouts)]
-    (log/infof "issue %s: confirmed payout: %s" issue-id payout-hash)
+(defn update-payout-receipt [owner repo comment-id balance-eth tokens payee-login issue-id confirm-hash payout-hash contract-address updated]
```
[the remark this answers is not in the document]
rcullito: These pre/post conditions are pretty neat. They seem lighter weight than using `spec`, although would be curious to discuss more when one might be preferred over the other. In the linked example it seems that constraining via pre/post is nice for wrapping multiple functions that require those same assertions... 

Whereas my own feeling is that in the case of defining a single function and/or data structure a spec schema may be more straightforward. 

In any event, good to validate I agree! 🙏

src/cljs/commiteth/manage_payouts.cljs
```diff
@@ -304,9 +304,9 @@
 (defn count-pill [n]
   [:span.v-top.ml3.ph3.pv1.bg-black-05.gray.br3.f7 n])
 
-(defn salute [name]
```
rcullito: @martinklepsch just wanted to verify that this wasn't intended for defaulting or something? and can be safely removed?

martinklepsch: Oh whoops, seems I messed up some merge conflict. It should take the users name from the DB (instead of calling everybody `Andy` 😅  — will clean that up.)

rcullito: lol, cool. I think it does end up doing that and in this case "andy" is discarded. just wanted to verify. I can leave the removal in this current PR if that's easier and then you won't have to do anything :)

rcullito:
@asemiankevich thank you for looking at it so promptly! I hadn't been experiencing that but will double check now on my local branch to see if I introduced a regression that might be causing this.

rcullito:
@asemiankevich I've pushed out a fix for the multiple popups (nice catch on your part 🙏 ). In terms of what happens next, a green pending banner will be displayed at the top of the page. So I think it would be just out of view in the screenshot you posted. Although I can certainly examine if that's not happening for you on this most recent version.

churik:
## 1. Revocation message is not visible when user focus is not on the page top
Steps:
- create issue with bounty, fund it
- go to Manage Payouts
- scroll to Unclaimed bounties section
- click on three dots -> "Revoke"

### Actual result:
"Transaction sent. Refund is pending until the transaction is completed. Check its status here. " - is not visible for user unless he will scroll to the top

### Expected result:
The message is visible and placed in screen area (something like popup window)
**Screen:** 
scrol


**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

churik:
## 1. Revoked bounties are displayed in "Paid" section
Steps:
- create issue with bounty, fund it
- go to Manage Payouts
- click on three dots -> "Revoke"
- confirm transaction from MetaMask
- wait until transaction is completed

### Actual result:
bounty is displayed in "Paid" section

### Expected result:
Separate section for revoked bounties
**Screen:** 
revok

**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

churik:
## 3. "Revoke" option disappears and appears needlessly
 
Steps:
1)  create issue with bounty, fund it
2) go to Manage Payouts
3) click on three dots -> "Revoke". Pay attention to other unclaimed and funded bounties.
4) refresh the page. Pay attention to revoked bounty.

### Actual result:
3) "Revoke" option disappears on all bounties
4) "Revoke" option appear everywhere, even on already revoked bounty

### Expected result:
3) "Revoke" option disappears on revoked bounty only
4) No changes after reload

**Video:** http://take.ms/9tzIm
**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

churik:
## 4. Informational message about sending transaction doesn't contain info about requirement to confirm transaction from MetaMask when bot transaction will be finished
 
Steps:
1)  create issue with bounty, fund it
2) go to Manage Payouts
3) click on three dots -> "Revoke". Pay attention to message "Transaction sent. Refund is pending until the transaction is completed. Check its status here."

### Actual result:
message

### Expected result:
Should be obvious that for revoking bounty (I think the good idea is even displayed id or issue title of bounty in message) user SHOULD confirm transaction after one or several (depending on tokens variety in bounty) transactions will be completed.
IMO message should be more informative, i.e.:
**Transaction(s) for revoking #X issue are sent. Refund is pending until the transaction(s) are completed (check its status here), then final transaction will be sent to your MetaMask wallet. Confirm it to finish revocation.**

**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

churik:
## 5. When revocation failed buuton "revoke" is still available
 
Steps:
1)  create issue with bounty, fund it
2) go to Manage Payouts
3) click on three dots -> "Revoke". 
4) in MetaMask reject transaction
5)  click on three dots -> "Revoke" again

### Actual result:
bot transaction is generated;
no transaction to confirm in MetaMask;

### Expected result:
after revocation failed "Revoke" option shouldn't be available because transaction for payment will not appear anymore

**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

churik:
## 6. Sometimes issue stays in Unclaimed bounties after revocation
 
I couldn't detect specific requirements, faced with this behavior only once.
I can see that all transactions are successful, but  bounty is still placed in Unclaimed section, GH comment is not updated (after revocation 2 hours passed)

### Actual result:
revoked bounty in "Unclaimed"
ghghg

### Expected result:
bounty is displayed along with other revoked bounties
**Issue:** https://github.com/TestingEnvENV/TestEnvAtt5_1/issues/254
**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64
**logs:** [log.log](https://github.com/status-im/open-bounty/files/1995330/log1.log)

rcullito:
@churik thanks again for the feedback, I think I have a pretty good feel for the issues you have raised.

Individual comments & progress are inline above among the checkboxes, but I had two main questions:
 
1.) do you know the cause of the failure during your testing? was it rejecting via metamask? or navigating away from the dashboard? or something completely different?
2.) If not in the "Paid" section, where were you expecting these bounties to go, into a dedicated revoked section or not be displayed at all? @EugeOrtiz this is also something I had meant to circle up with you on but... well, here we are :) 

Thanks both! - Rob

rcullito:
@churik feedback addressed in line above ^^. pretty happy with where this landed and should be good for another round of testing 🙏

churik:
## 9. GH usernames for revoked bounties in list and in issue don't match

Steps:
1)  create issue with bounty, fund it
2) go to Manage Payouts
3) click on three dots -> "Revoke".
4) wait until transaction is completed
5) pay attention to "Paid out to" value in "Paid out" section
6) open issue on GitHub, pay attention to bot comment "Paid out" value

### Actual result:
values are different
po1
po2

### Expected result:
values should be the same in GH comment and in "Paid out" SOB section

**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64

churik:
## 10.  After updating payment address address in informational message is not updated automatically

Steps:
1)  create issue with bounty, fund it
2) go to Manage Payouts
3) click on three dots -> "Revoke".
4) go to My payment details
5) set new address and click on "Save"
6) repeat steps 2-3

### Actual result:
value is updated onlya fter manual reload
unt

### Expected result:
value is updated automatically

**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64
**Video:** http://take.ms/5oCUD
image

churik:
## 11.  No transaction in MetaMask for revocation confirming

With last changes issue appered that transaction is not triggered when bot transaction is already completed, so revocation is broken.

Steps:
1)  create an issue with bounty, fund it
2) go to Manage Payouts
3) click on three dots -> "Revoke".
4) wait until transaction is comleted
5) wait for 10 min more (for MetaMask confirmation window)

### Actual result:
no confirmation windoe in MetaMask

### Expected result:
confirmation window appears once bot transaction is completed
**Issues:**
https://github.com/TestingEnvENV/TestEnvAtt5_1/issues/264
https://github.com/TestingEnvENV/TestEnvAtt5_1/issues/263
https://github.com/TestingEnvENV/TestEnvAtt5_1/issues/261

**Wallet**: MetaMask
**OS:**  Mac OS High Sierra
**Browser:** Chrome 64
**Log:**
[log1.log](https://github.com/status-im/open-bounty/files/2013058/log1.log)

churik:
@rcullito thanks for your work!

unfortunately, I still don’t see any transactions in MetaMask for revocation confirming. So I’m still blocked with testing 
https://github.com/status-im/open-bounty/pull/430
I don’t know where exactly issue could be - probably it is even not our part, but 
1) I didn’t upgrade MetaMask
2) all that I have done - change account in My payment details several times (I did it several times before last changes in PR)
All that I’ve got now - bot transaction is completed, but no more transaction is triggered through MetaMask so revocation can’t proceed.
For first time issue reproduced during testing iteration 16-17.05.18.
Before these changes, this behavior wasn't reproducible

Also, please take a look into this comment:

>  "Revoke" option disappears and appears needlessly
> (RC) - the intended behavior is that the option will not be visible during a revocation. This is the behavior I have been experiencing on successful revokes. Although I'll keep an eye on this while addressing your other points about funky state during failed revocations.
> @churik: I've experiencing behavior when issue state and availability of "Revoke" button change when you manually updated the page. I don't think it is OK (video: http://take.ms/vIGbH)

I'm still experiencing with this behavior.

rcullito:
Ah ok, the video helps! Thanks for that. The difference in our experiences I believe is that I am not refreshing during my own process, but that is a fair point to bring up. I'll push some changes to persist our app state for revocation into localStorage so that it is not as brittle.

rcullito:
@churik alright this should be pretty bullet proof at this point, resistant to page refresh, earthquakes... whatever else might come up lol. 

Thank you for your help! This should be much more robust now if you are able to test again. Please let me know if you are not experiencing an improvement.

churik:
To summarize for now:
1) After updating payment address it is not updated automatically in the informational message - [existing issue](https://github.com/status-im/open-bounty/pull/430#issuecomment-389797906)
2) Revoked bounties are not displayed separately on dashboard, but they don't affect on "Paid for 1 solved bounties"

PR #435, opened 2018-04-25
fix #434: dont fetch price-info too often

### LINKED ISSUE #434: Reduce amount of requests sent to FIAT value providers
In `commiteth.util.crypto-fiat-value` we define a bunch of utility functions to retrieve the fiat value of various cryptocurrencies.  As part of `commiteth.scheduler/update-balances` these functions are called for **every bounty every 10 minutes**. 

This results in a lot of requests to our fiat value provider (coinmarketcap usually) causing them to shut us off with `429 Too Many Requests`.

We should implement a mechanism to cache our fiat value provider's response for a limited amount of time and only refresh this data if it's too old.

martinklepsch:
**QA NOTES** This change has no user facing effect. It changes how we fetch
price data for our background jobs. Previously everytime we wanted to know the
USD equivalent of some bounties tokens we made a request to some API.

This caused us to get rate limited as described in #434.

This change modifies our code so that we cache the conversion rate for
5 minutes. If you watch the logs on testing you should see log messages like

```
:SNT: Getting price-data from CoinMarketCap (token-id status)
```
appearing once every 10 minutes or so. You should not see these messages
in quick succession.

### DIFF
--- a/src/clj/commiteth/util/crypto_fiat_value.clj
+++ b/src/clj/commiteth/util/crypto_fiat_value.clj
@@ -1,66 +1,100 @@
 (ns commiteth.util.crypto-fiat-value
   (:require [mount.core :as mount]
+            [clj-time.core :as t]
+            [clojure.edn :as edn]
             [clojure.tools.logging :as log]
             [commiteth.config :refer [env]]
             [commiteth.util.util :refer [json-api-request]]))
 
 
-(defn fiat-api-provider []
+(defn- fiat-api-provider []
   (env :fiat-api-provider :coinmarketcap))
 
-
-(defn get-token-usd-price-cryptonator
+(defn- get-token-usd-price-cryptonator
   "Get current USD value for a token using cryptonator API"
   [tla]
+  (log/infof "%s: Getting price-data from Cryptonator" tla)
   (let [token (subs (str tla) 1)
-        url (str "https://api.cryptonator.com/api/ticker/"
-                 token
-                 "-usd")
+        url (format "https://api.cryptonator.com/api/ticker/%s-usd" token)
         m (json-api-request url)]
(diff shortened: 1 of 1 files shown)


siphiuel:
Why do we use `edn/read-string` instead of `read-string`?

martinklepsch:
@siphiuel 
```clojure
(read-string "#=(clojure.java.shell/sh \"echo\" \"hi\")")
```

siphiuel:
@martinklepsch I see, thanks.

src/clj/commiteth/util/crypto_fiat_value.clj
```diff
@@ -1,66 +1,100 @@
...
+  "Get current USD value for the crypto-amounts passed as argument.
+  Example: {:ETH 123, :SNT 456}"
+  [crypto-amounts]
+  (->> crypto-amounts
+       (map (fn [[tla value]]
+              (convert-usd fiat-converter tla value)))
+       (reduce + 0)))
+
+
+(comment
```
rcullito: had been meaning to ask about the use of these. are they meant to indicate how one might easily test this functionality if the comment was removed?

martinklepsch: I often use these to keep REPL snippets related to that namespace around. Usually I move my curser after the form I want to evaluate and evaluate the preceding form (`C-x C-e` in Emacs I think). This way I can easily run smaller functions and figure out if they're doing what they're supposed to :)

rcullito: Makes total sense! Thanks for sharing.

churik:
### Environments

### Tested functionality
* Contract deployment;
* Bounty funding - info updated on "Bounties", in GH bot comment

PR #456, opened 2018-05-24
store currently confirming payouts in local storage

rcullito:
The main change here is that the `confirming?` flag is being persisted in local storage, as before there was no record of whether a confirm payout had been initiated if the page was refreshed. 

Additionally the confirm button itself will be hidden during confirmation. Previously it was set to a disabled state that was not discernibly different different from the active state, which may have been the source of confusion. 

So to be clear this won't materially guarantee that the payout hash is set, but it should make it easier to spot if the problem happens again, and users won't even be given the option to confirm the payout multiple times.

### DIFF
--- a/src/cljs/commiteth/handlers.cljs
+++ b/src/cljs/commiteth/handlers.cljs
@@ -425,6 +425,14 @@
 (defn strip-0x [x]
   (str/replace x #"^0x" ""))
 
+(reg-event-fx
+ :set-confirming-payout
+ [interceptors (inject-cofx :store)]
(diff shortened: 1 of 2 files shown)


rcullito:
@siphiuel yes, as far as I could tell the web3 js api that we are using doesn't support the same .getLogs method that I used before over the json rpc on the back end. So there didn't look like a meaningful way to update or check the status of a contract's confirmation beyond what we are currently doing: the best we can do is try and clearly communicate that we haven't "heard back" one way or the other by hiding the button.

